PR #1, opened 2023-02-14
Map provider to forked provider name when different

aq17:
This PR implements a few minor changes:
- Usually the pulumi provider name is the same as the upstream, but we need to map it correctly when it differs (i.e. `pulumi-confluentcloud` to `confluent`)
- For forked providers, fetches all tags of upstream provider so the latest can be merged into a new branch of the forked provider


iwahbe (changes requested):
We should never swallow errors. I would prefer it if we check for a clean build tree before committing the `make build_sdks` step. If the work tree is clean, we can run `git commit -m "make build_sdks: no changes" --allow-empty`.

iwahbe (approved):

PR #6, opened 2023-02-16
Set GOWORK=off

iwahbe:
This adds a "Setting Up Environment" step, which we nest this under. This allows a user running the command to understand what the cli is doing under the hood.


aq17 (approved):
😄

PR #19, opened 2023-03-16
Detect NoOp bridge and provider updates

iwahbe:
`upgrade-provider` will no detect the current version of the bridge before it attempts to upgrade the bridge. If the current and target bridge version match, it will not attempt the update.

This also PR changes the behavior when attempting to update a provider version, but no version is detected. Instead of erroring, it succeeds.

If no action is needed, `upgrade-provider` will stop at the discovery step:

```
$ upgrade-provider pulumi-databricks
---- Setting Up Environment ----
- ✓ GOWORK="off": done
- ✓ PULUMI_MISSING_DOCS_ERROR="true": done
---- Discovering Repository ----
- Ensure 'github.com/pulumi/pulumi-databricks'
  - ✓ Expected Location: /Users/ianwahbe/go/src/github.com/pulumi/pulumi-databricks
  - ✓ Downloading: skipped
  - ✓ Validating: done
- pull default branch
  - ✓ /usr/local/bin/git ls-remote --heads origin: done
  - ✓ finding default branch: main
  - ✓ /usr/local/bin/git checkout main: done
  - ✓ /usr/local/bin/git pull origin: done
- ✓ Target Provider Version: Up to date
- ✓ Repo kind: plain
- ✓ Planning Bridge Update: Up to date at v3.42.1
No actions needed (in bold, but it doesn't appear you can bold text within a markdown block)
```

Normally `upgrade-provider pulumi-databricks` would attempt both a provider and bridge update.

---

This will enable firing `upgrade-provider --kind=bridge pulumi-${PROVIDER}` in a loop against all of our providers. Providers that need an update will get one. Providers that don't need an update won't have a PR set up.

### DIFF
--- a/colorize/colorize.go
+++ b/colorize/colorize.go
@@ -1,0 +1,16 @@
+package colorize
+
+const (
+	esc   = "\u001B["
+	bold  = esc + "1m"
+	warn  = bold + esc + "33m"
+	reset = esc + "m"
+)
+
+func Bold(s string) string {
+	return bold + s + reset
+}
+
+func Warn(s string) string {
+	return warn + s + reset
+}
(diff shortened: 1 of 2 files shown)


aq17 (approved):
LGTM!

PR #20, opened 2023-03-16
Set up tag based release workflow

iwahbe:
I'm doing this so we can version downstream CI consumers (when we have downstream CI consumers).

The contents of the action is based on [schema-tools](https://github.com/pulumi/schema-tools/blob/e3106cb65dbde7c8f0646ca4110600415bdd9673/.goreleaser.yml#L1-L25).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,24 @@
+name: release
+on:
+  push:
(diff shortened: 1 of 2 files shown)


aq17 (approved):
👍🏼

PR #22, opened 2023-03-20
Document previous upstream version of a patched provider

iwahbe:
This only works for patched providers right now, but I plan to extend the functionality to all providers (where possible). 

This is prerequisite work for determining the correct tag increment for an upgrade. (v2.2.1 -> v3.0.0 is a major update, v3.2.1 -> v3.3.1 is a minor update, v3.2.1 -> v3.2.2 is a patch update).


aq17 (approved):
👍🏼

PR #26, opened 2023-03-27
Detect previous upstream versions (continued)

iwahbe:
Ensure that we have detected the previous upstream provider version correctly in all cases.

We already handled plain and patched providers. This PR adds forked and shimmed support, bringing upstream version detection to full capability.


aq17 (approved):
LGTM

PR #30, opened 2023-04-03
Upgrade TF Plugin SDK to latest fork on provider upgrade

### LINKED ISSUE #27: Keep the pinned version of the TF Plugin SDK up to date
All of our providers have a line such as

```
replace (
	github.com/hashicorp/terraform-plugin-sdk/v2 => github.com/pulumi/terraform-plugin-sdk/v2 v2.0.0-20230327102345-3fa930f86570
	...
)
```

When the upstream providers require a new version of the TF Plugin SDK, `upgrade-provider` fails. 

We should automate keeping the version in the replace the same as the version in the upstream repo. Upgrading from `v2.24.1` to `v2.26.1` should have a mirrored upgrade in the `replace` directive.

iwahbe:
Fixes #27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  GOLANGCI_LINT_VERSION: v1.50.1
+  GOLANGCI_LINT_VERSION: v1.52.2
 name: Pull Request
 on: [pull_request]
 jobs:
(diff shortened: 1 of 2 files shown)


jazzyfresh (approved):
A couple things
- I know that the current testing process is to run this on an existing provider and iterate, but that's almost like testing in production (or at least, local staging). We should really get some sort of unit testing framework for this project in place. We can't release this as a production ready tool without a test suite.
- We're getting to the point that file is getting huge and we should refactor into multiple files for readability. However, this isn't the largest Go file I've ever seen, so I suppose this is fine for now. Once we hit 10,000 lines we should break it up though!

aq17 (approved):

PR #39, opened 2023-04-14
Add AutoAliasing code migration

aq17:
This PR adds an option to perform code migrations on providers, either alongside a normal provider updates (by default or if --kind=all), or by itself (if --kind=code).
If --kind=all, all available code migrations options will be performed.
If --kind=code, a list of options must be specified via --migration-opts.

Currently the only option is to add AutoAliasing.


iwahbe (approved):
A couple nits, but otherwise LGTM

PR #46, opened 2023-04-21
Use current directory if targeted repo

### LINKED ISSUE #42: Support running upgrade provider from an already cloned repository
Enable running upgrade-provider against a repository already cloned, e.g.: locally or via `@actions/checkout`.

aq17:
Fixes https://github.com/pulumi/upgrade-provider/issues/42
If we run `upgrade-provider pulumi-{provider}` from within `pulumi-{provider}`, we should default to using the current repo instead of cloning

### DIFF
--- a/upgrade/steps-helpers_test.go
+++ b/upgrade/steps-helpers_test.go
@@ -1,6 +1,7 @@
 package upgrade
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


upgrade/steps.go
```diff
@@ -310,7 +310,6 @@ func InformGitHub(
 		panic("Unknown action")
 	}
 	createPR := step.Cmd(exec.CommandContext(ctx, "gh", "pr", "create",
-		"--assignee", "@me",
```
aq17, line 301: For CI, errors on `could not find user pulumi-bot`. Removing for now to test

upgrade/steps.go
```diff
@@ -298,7 +308,7 @@ func InformGitHub(
...
+		"--add-assignee", "@me",
```
iwahbe, line 311: Why this change?

upgrade/steps.go
```diff
@@ -140,7 +140,17 @@ func ensureUpstreamRepo(ctx Context, repoPath string) step.Step {
...
+			if err != nil {
+				return "", fmt.Errorf("could not resolve current directory: '%s", err)
+			}
+			if idx := strings.Index(cwd, filepath.Base(repoPath)); idx != -1 {
+				idx = idx + len(expectedLocation) - 1
+				expectedLocation = cwd[:idx]
+			} else {
+				expectedLocation = filepath.Join(ctx.GoPath, "src", expectedLocation)
+
+			}
```
iwahbe, line 153: Can you explain what this is trying to do. It seems like the logic should be
```go
if filepath.Base(repoPath) == filepath.Base(cwd) && filepath.Base(cwd) != "" {
    	expectedLocation = cwd
else {
    	expectedLocation = filepath.Join(ctx.GoPath, "src", expectedLocation)
}
```

aq17: I think I was trying to account for the possibility of being in a subdirectory, i.e. `user/home/pulumi/pulumi-docker/provider`would set the expected location to `user/home/pulumi/pulumi-docker`

upgrade/steps.go
aq17: Accident :(

upgrade/steps-helpers_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetRepoExpectedLocation(t *testing.T) {
+	ctx := Context{
+		GoPath: "Users/myuser/go",
+	}
+
+	cwd, err := os.Getwd()
+	assert.Nil(t, err)
```
iwahbe, line 18: Is there a way we can do this so that our tests don't depend on the current working directory? We could make `getRepoExpectedLocation` a pure function: `func(goPath, cwd, repoPath string) string`.

upgrade/steps-helpers_test.go
```diff
@@ -0,0 +1,48 @@
...
+	mockRepoPath = trimSeparators(mockRepoPath)
+	expected = trimSeparators(expected)
+	assert.Equal(t, mockRepoPath, expected)
+
+	// test cwd completely different from repo path
+	mockRepoPath = filepath.Join("github.com", "pulumi", "random-provider")
+	expected, err = getRepoExpectedLocation(ctx, mockRepoPath)
+	expected = trimSeparators(expected)
+	assert.Nil(t, err)
+	assert.Equal(t, filepath.Join(ctx.GoPath, "src", mockRepoPath), expected)
```
iwahbe, line 41: This pattern is perfect for table based testing. That would look something like this:
```suggestion
	tests := []struct{ repoPath, expected string }{
		{ cwd, trimSeparators(expected) },
		{ strings.TrimSuffix(cwd, filepath.Base(cwd)), getRepoExpectedLocation(ctx, mockRepoPath) },
		{ filepath.Join("github.com", "pulumi", "random-provider"), trimSeparators(expected) },
	}
	
	for _, tt := range tests {
		tt := tt
		t.Run(fmt.Sprintf("(%s,%s)",tt.repoPath,tt.expected), func(t *testing.T) {
			mockRepoPath = filepath.Join("github.com", "pulumi", "random-provider")
			expected, err = getRepoExpectedLocation(ctx, mockRepoPath)
			expected = trimSeparators(expected)
			assert.Nil(t, err)
			assert.Equal(t, tt.expected, expected)
		})
	// test cwd completely different from repo path

```

aq17: Thanks for the suggestion, applied

iwahbe (approved):
Thanks for the work. I think this is much cleaner. LGTM!

PR #60, opened 2023-04-25
User option: allow multiple kinds

iwahbe:
This is rebased on top of https://github.com/pulumi/upgrade-provider/pull/59 so I can do working tests, but the change is separate. If we merge it after #59 it will show up as a fully separate commit into main.


aq17 (approved):
LGTM 👍🏼

Issue #55, opened 2023-04-25
Allow any upstream repo target path


aq17:
Currently, `upgrade-provider` takes in the name of a provider repo and expects in to be in the `pulumi` org. We should extend this to target any org/ repo (which should work if proper git creds are set up)

aq17:
Duplicate of https://github.com/pulumi/upgrade-provider/issues/65

PR #61, opened 2023-04-26
Accept config file

### LINKED ISSUE #69: Add experimental flag
Add experimental flag/ config value to explicitly set whether experimental bridge features should be enabled

aq17:
Fixes https://github.com/pulumi/upgrade-provider/issues/527
Fixes #69 

This PR adds the option to read a config file `upgrade-config.{yml/json}` in a provider repo instead of passing in values as flags to `upgrade-provider`. 
The name of the value should match the name of the flag (i.e. `--upstream-provider-name` as flag, `upstream-provider-name` as config value).
Passing in a flag should take precedence over setting the config value

It also adds in an additional flag `upstream-provider-name` as part of https://github.com/pulumi/upgrade-provider/issues/55, and an additional flag `experimental` as part of https://github.com/pulumi/upgrade-provider/issues/69. Both are good examples to set in the config file instead of flag


main.go
```diff
@@ -102,10 +117,76 @@ If the passed version does not exist, an error is signaled.`)
...
+
+	return nil
+}
+
+// Bind each cobra flag to its associated viper configuration (config file and environment variable)
+func bindFlags(cmd *cobra.Command, v *viper.Viper) {
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		// Determine the naming convention of the flags when represented in the config file
+		configName := f.Name
+		// If using camelCase in the config file, replace hyphens with a camelCased string.
```
guineveresaenger, line 178: should this read:
```suggestion
		// If using hyphenated names in the config file, replace hyphens with a camelCased string.
```
?

main.go
```diff
@@ -102,10 +117,76 @@ If the passed version does not exist, an error is signaled.`)
...
+		// Since viper does case-insensitive comparisons, we don't need to bother fixing the case, and only need to remove the hyphens.
+		if replaceHyphenWithCamelCase {
```
guineveresaenger, line 180: where and why does this get set to `true`?

main.go
```diff
@@ -102,10 +117,76 @@ If the passed version does not exist, an error is signaled.`)
...
+			configName = strings.ReplaceAll(f.Name, "-", "")
```
guineveresaenger, line 181: I'm probably missing some context here! this doesn't seem to camelCase anything?

main.go
```diff
@@ -102,10 +117,76 @@ If the passed version does not exist, an error is signaled.`)
...
+	if err := v.ReadInConfig(); err != nil {
+		// It's okay if there isn't a config file
+		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
+			return err
+		}
+	}
+
+	// When we bind flags to environment variables expect that the
+	// environment variables are prefixed, e.g. a flag like --number
+	// binds to an environment variable STING_NUMBER. This helps
```
guineveresaenger, line 154:
```suggestion
	// binds to an environment variable UPGRADE_NUMBER. This helps
```

main.go
```diff
@@ -102,10 +117,76 @@ If the passed version does not exist, an error is signaled.`)
...
+		configName := f.Name
+		// If using camelCase in the config file, replace hyphens with a camelCased string.
+		// Since viper does case-insensitive comparisons, we don't need to bother fixing the case, and only need to remove the hyphens.
+		if replaceHyphenWithCamelCase {
+			configName = strings.ReplaceAll(f.Name, "-", "")
+		}
+		fmt.Println(configName)
+
+		// Apply the viper config value to the flag when the flag is not set and viper has a value
+		if !f.Changed && v.IsSet(configName) {
```
guineveresaenger, line 186: Does this mean flags override config settings?

aq17: Yes, I think flags should override config

main.go
aq17: It's not really `replaceHyphenWithCamelCase`, it's really just `removeHyphens` which functionally does the same thing in viper. I'll change the name to be clearer

main.go
aq17: It just depends if we want to be lenient with the format our users must use for config values (i.e. `config-val-one` and`configValOne` both set the same config), or if we want to be strict and require one way. Either approach is fine with me

guineveresaenger: Hmmm, honestly I think it's fine to allow both. This answers my why, but I'm not seeing the logic where we parse for hyphens and set it to true.

main.go
aq17: Honestly I think it's easier to be strict and and require with the config value format to match the flag, i.e. (`upstream-provider-name` is valid, `upstream_provider_name` or `upstreamProviderName` are not`). I'll change it to that for now

upgrade/steps-helpers.go
```diff
@@ -139,13 +139,6 @@ func say(msg string) func([]byte) (string, error) {
 	}
 }
 
-func getTfProviderRepoName(providerName string) string {
```
aq17, line 142: Instead of relying on a manual mapping of pulumi provider names to the Tf provider name, or assuming they are equivalent, it should be set in the config file (or with `--upstream-provider-name`)

aq17:
[Example PR](https://github.com/pulumi/pulumi-oci/pull/118/files#diff-768cb5c56f5bda1bf97a6d9736aaaf6cb7743afeb3162ef1baff9a5434b9b97c) with `upstream-provider-name` set in a config file

upgrade/util.go
```diff
@@ -47,6 +49,8 @@ type ProviderRepo struct {
 	// The upstream version we are upgrading from.  Because not all upstream providers
 	// are go module compliment, we might not be able to always resolve this version.
 	currentUpstreamVersion *semver.Version
+
+	upstreamName string
```
iwahbe, line 53: Is this different than `Context.UpstreamProviderName`? Can you add a doc string explaining what it means?

main.go
```diff
@@ -6,15 +6,26 @@ import (
...
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
 
 	"github.com/pulumi/upgrade-provider/colorize"
 	"github.com/pulumi/upgrade-provider/upgrade"
 )
 
+var (
+	// The name of our config file, without the file extension because viper supports many different config file languages.
+	defaultConfigFilename = "upgrade-config"
```
iwahbe, line 23: I thought we discussed that this should start with a `.` so it is hidden.

Also, why is it `defaultConfigName`? Can you override the config name? If so, where?

iwahbe (commented):
We need to add some documentation to the README explaining what options are supported and where a config file should live. At the very least, we should have a fully annotated example `upgrade-config.yaml` in a code block in `README.md`.

upgrade/util.go
aq17: They are the same, but not necessary to keep track of it in `ProviderRepo`. Simpler to store it in the context only so I'll remove

iwahbe (approved):

PR #64, opened 2023-05-01
Error if target version < current version

### LINKED ISSUE #63: Make `upgrade-provider` robust to prior versions
If a provider upgrade attempts to "upgrade" to a lower version than what it is currently using, we should error out

aq17:
Fixes #63 

If a target upgrade version is less than the current version, the update should error.

### DIFF
--- a/upgrade/upgrade-provider.go
+++ b/upgrade/upgrade-provider.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"golang.org/x/mod/semver"
 
 	"github.com/pulumi/upgrade-provider/colorize"
 	"github.com/pulumi/upgrade-provider/step"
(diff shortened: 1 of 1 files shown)


upgrade/upgrade-provider.go
```diff
@@ -89,6 +90,11 @@ func UpgradeProvider(ctx Context, name string) error {
 
 				var previous string
 				if repo.currentUpstreamVersion != nil {
+					if semver.Compare(repo.currentUpstreamVersion.String(),
+						upgradeTargets.Latest().String()) > 0 {
```
guineveresaenger, line 94: I wonder if we could/should filter these out in [GetExpectedTargets](https://github.com/pulumi/upgrade-provider/blob/main/upgrade/steps.go#L496)?

upgrade/upgrade-provider.go
```diff
@@ -89,6 +90,11 @@ func UpgradeProvider(ctx Context, name string) error {
...
+						return "", fmt.Errorf("current upstream version %v is greater than target version %v",
```
guineveresaenger, line 95: sweet error message 👍

upgrade/upgrade-provider.go
aq17: `GetExpectedTargets` fetches the latest version (<= `--provider-version` if set), and then we get the current version from `setCurrentUpstream*` 
I think we can keep it here for now, but I think we should add a separate function to directly fetch the upstream version vs looking at issues?

guineveresaenger: I'm fine keeping it here, you make a good point about needing a separate function that bypasses the issue crawling.

guineveresaenger (approved):
If we need to refine semver comparison in the future, we can take a look at [Masterminds semver package](github.com/Masterminds/semver), as used [here](https://github.com/pulumi/release-bot/pull/41) but the basic package is plenty sufficient here.

PR #66, opened 2023-05-01
Accept {org/repo} as argument to upgrade-provider

### LINKED ISSUE #65: Accept {org/repo} as argument to upgrade-provider
Instead of taking a repository name and assuming it is in the `pulumi` org (in GitHub), we should accept any org/repo name

aq17:
Fixes #65 
Instead of assuming the target repo is part of the pulumi github org, we require the org/repo name to be passed in

### DIFF
--- a/main.go
+++ b/main.go
@@ -59,6 +59,11 @@ func cmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			// Require `upstream-provider-name` to be set
+			if context.UpstreamProviderName == "" {
+				return errors.New("`upstream-provider-name` must be provided")
+			}
+
 			// Validate that maxVersion is a valid version
 			if maxVersion != "" {
 				context.MaxVersion, err = semver.NewVersion(maxVersion)
--- a/main.go
+++ b/main.go
@@ -153,8 +153,7 @@ If the passed version does not exist, an error is signaled.`)
 - "autoalias": Apply auto aliasing to the provider.`)
 
 	cmd.PersistentFlags().StringVar(&context.UpstreamProviderName, "upstream-provider-name", "",
-		`The name of the upstream provider.
-Must be set in '.upgrade-config.yml' or passed with flag.`)
+		`The name of the upstream provider.`)
 
 	return cmd
 }
--- a/main.go
+++ b/main.go
@@ -34,6 +34,8 @@ func cmd() *cobra.Command {
 	}
 	var upgradeKind []string
 	var experimental bool
+	var repoName string
+	var repoOrg string
 
 	context := upgrade.Context{
 		Context: context.Background(),
@@ -54,11 +56,18 @@ func cmd() *cobra.Command {
 		Use:   "upgrade-provider",
 		Short: "upgrade-provider automatics the process of upgrading a TF-bridged provider",
 		Args:  cobra.ExactArgs(1),
-		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			err := initializeConfig(cmd)
 			if err != nil {
 				return err
 			}
+			// Validate argument is {org}/{repo}
+			tok := strings.Split(args[0], "/")
+			if len(tok) != 2 {
+				fmt.Println("error: argument must be provided as {org}/{repo}")
+				os.Exit(1)
+			}
+			repoOrg, repoName = tok[0], tok[1]
 			// Require `upstream-provider-name` to be set
 			if context.UpstreamProviderName == "" {
 				return errors.New("`upstream-provider-name` must be provided")
@@ -117,13 +126,7 @@ func cmd() *cobra.Command {
 			return nil
 		},
 		Run: func(_ *cobra.Command, args []string) {
-			// Validate argument is {org}/{repo}
-			tok := strings.Split(args[0], "/")
-			if len(tok) != 2 {
-				fmt.Println("error: argument must be provided as {org}/{repo}")
-				os.Exit(1)
-			}
-			err := upgrade.UpgradeProvider(context, tok[0], tok[1])
+			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
 			exitOnError(err)
 		},
 	}
--- a/main.go
+++ b/main.go
@@ -146,7 +146,8 @@ If the passed version does not exist, an error is signaled.`)
 - "autoalias": Apply auto aliasing to the provider.`)
 
 	cmd.PersistentFlags().StringVar(&context.UpstreamProviderName, "upstream-provider-name", "",
-		"The name of the upstream provider.")
+		`The name of the upstream provider.
+Must be set in '.upgrade-config.yml' or passed with flag.`)
 
 	return cmd
 }
--- a/main.go
+++ b/main.go
@@ -64,8 +64,7 @@ func cmd() *cobra.Command {
 			// Validate argument is {org}/{repo}
 			tok := strings.Split(args[0], "/")
 			if len(tok) != 2 {
-				fmt.Println("error: argument must be provided as {org}/{repo}")
-				os.Exit(1)
+				return errors.New("argument must be provided as {org}/{repo}")
 			}
 			repoOrg, repoName = tok[0], tok[1]
 			// Require `upstream-provider-name` to be set
--- a/main.go
+++ b/main.go
@@ -67,6 +67,10 @@ func cmd() *cobra.Command {
 				return errors.New("argument must be provided as {org}/{repo}")
 			}
 			repoOrg, repoName = tok[0], tok[1]
+			// repo name should start with 'pulumi-'
+			if !strings.HasPrefix(repoName, "pulumi-") {
+				return errors.New("{repo} must start with `pulumi-`")
+			}
 			// Require `upstream-provider-name` to be set
 			if context.UpstreamProviderName == "" {
 				return errors.New("`upstream-provider-name` must be provided")
--- a/main.go
+++ b/main.go
@@ -112,7 +112,13 @@ func cmd() *cobra.Command {
 			return nil
 		},
 		Run: func(_ *cobra.Command, args []string) {
-			err := upgrade.UpgradeProvider(context, args[0])
+			// Validate argument is {org}/{repo}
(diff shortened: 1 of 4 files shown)


aq17:
Example upgrade https://github.com/pulumi/pulumi-mongodbatlas/pull/214

upgrade/steps.go
```diff
@@ -845,8 +845,8 @@ func migrationSteps(ctx Context, repo ProviderRepo, providerName string, descrip
 	return steps, nil
 }
 
-func AddAutoAliasing(ctx Context, repo ProviderRepo, providerName string) (step.Step, error) {
-	providerName = strings.TrimPrefix(providerName, "pulumi-")
+func AddAutoAliasing(ctx Context, repo ProviderRepo) (step.Step, error) {
+	providerName := strings.TrimPrefix(repo.name, "pulumi-")
```
iwahbe, line 849: It looks like we are still assuming that the provider's name will start with `pulumi-`, even though we let the user pass in information to the contrary: `iwahbe/my-pulumi-provider` as the repo name.

upgrade/upgrade-provider.go
```diff
@@ -27,7 +28,15 @@ func UpgradeProvider(ctx Context, name string) error {
...
-	upstreamProviderName := strings.TrimPrefix(name, "pulumi-")
+
+	tok := strings.Split(name, "/")
+	if len(tok) != 2 {
+		return errors.New("argument must be provided as {org}/{repo}")
+	}
+	repo.org = tok[0]
+	repo.name = tok[1]
+
+	upstreamProviderName := strings.TrimPrefix(repo.name, "pulumi-")
```
iwahbe, line 39: Should we verify that `repo.name` starts with `pulumi-`?

upgrade/upgrade-provider.go
```diff
@@ -27,7 +28,15 @@ func UpgradeProvider(ctx Context, name string) error {
 	var targetBridgeVersion string
 	var upgradeTargets UpstreamVersions
 	var goMod *GoMod
-	upstreamProviderName := strings.TrimPrefix(name, "pulumi-")
+
+	tok := strings.Split(name, "/")
+	if len(tok) != 2 {
+		return errors.New("argument must be provided as {org}/{repo}")
+	}
```
iwahbe, line 35: Can we move this validation to the part of the repo that handles CLI interaction. We can then change `UpgradeProvider` to take `(ctx Context, repoOrg, repoName string)`.

upgrade/upgrade-provider.go
aq17: I should have specified that this should be replaced by https://github.com/pulumi/upgrade-provider/pull/61. We shouldn't assume the upstream is prefixed with `pulumi`, it should be explicitly set in the config/ passed in with `--upstream-provider-name`

upgrade/steps.go
aq17: Rebased on main now

main.go
```diff
@@ -59,6 +59,11 @@ func cmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			// Require `upstream-provider-name` to be set
+			if context.UpstreamProviderName == "" {
+				return errors.New("`upstream-provider-name` must be provided")
+			}
```
iwahbe, line 65: If possible, can we do this within our CLI framework (cobra)?

If we are declaring this command required, should we make it positional?

main.go
```diff
@@ -112,7 +117,13 @@ func cmd() *cobra.Command {
 			return nil
 		},
 		Run: func(_ *cobra.Command, args []string) {
-			err := upgrade.UpgradeProvider(context, args[0])
+			// Validate argument is {org}/{repo}
+			tok := strings.Split(args[0], "/")
+			if len(tok) != 2 {
+				fmt.Println("error: argument must be provided as {org}/{repo}")
+				os.Exit(1)
+			}
```
iwahbe, line 125: This is what `PersistentPreRunE` is for.

upgrade/steps.go
iwahbe, line 849: I still don't understand. Provider name is a user option flag. We shouldn't need to trim `pulumi-` to get it. If this is us applying the default, let's apply it earlier and so consuming it here is identical to consuming the user option.

upgrade/upgrade-provider.go
```diff
@@ -13,20 +13,23 @@ import (
...
+func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 	var err error
 	var repo ProviderRepo
 	var targetBridgeVersion string
 	var upgradeTargets UpstreamVersions
 	var goMod *GoMod
 
+	repo.org = repoOrg
+	repo.name = repoName
+
```
iwahbe, line 32: Lets just make this part of the struct declaration for `repo`:
```go
repo := ProviderRepo {
   name: repoName,
   org: repoOrg,
}
```

upgrade/upgrade-provider.go
```diff
@@ -115,7 +118,7 @@ func UpgradeProvider(ctx Context, name string) error {
 		discoverSteps = append(discoverSteps,
 			step.F("Current Major Version", func() (string, error) {
 				var err error
-				repo.currentVersion, err = latestRelease(ctx, "pulumi/"+name)
+				repo.currentVersion, err = latestRelease(ctx, repoName)
```
iwahbe, line 121: I think this was the GH path, so `org/name`. Are you sure that this works?

iwahbe (changes requested):

main.go
```diff
@@ -42,7 +43,7 @@ func cmd() *cobra.Command {
 		Use:   "upgrade-provider",
 		Short: "upgrade-provider automatics the process of upgrading a TF-bridged provider",
 		Args:  cobra.ExactArgs(1),
-		PersistentPreRunE: func(_ *cobra.Command, _ []string) error {
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
```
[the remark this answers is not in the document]
aq17: To call `initializeConfig(cmd)`

upgrade/steps.go
aq17: I believe the user input is {org}/{repo}, so say `repo.org == `pulumi`, `repo.name` == `pulumi-xyz`. Here, we assume the subdirectory should be `pulumi-resource-xyz` containing the metadata file ([following the boilerplate](https://github.com/pulumi/pulumi-tf-provider-boilerplate/tree/main/provider/cmd), so we trim `pulumi-` from `repo.name`. 
Before, providerName was just the repo name so I think this should be equivalent?

aq17: Tested on https://github.com/pulumi/pulumi-pagerduty/pull/194

main.go
aq17: Good question, I'm not sure. I think it's better to specify it in the config file (which sets the corresponding flag), but if not, require the flag? Unless there's a benefit to having it positional

main.go
```diff
@@ -54,11 +56,18 @@ func cmd() *cobra.Command {
...
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			err := initializeConfig(cmd)
 			if err != nil {
 				return err
 			}
+			// Validate argument is {org}/{repo}
+			tok := strings.Split(args[0], "/")
+			if len(tok) != 2 {
+				fmt.Println("error: argument must be provided as {org}/{repo}")
+				os.Exit(1)
```
iwahbe, line 68:
```suggestion
				return errors.New("argument must be provided as {org}/{repo}")
```

main.go
iwahbe: We can leave it a flag. Positional just helps with brevity.

iwahbe: Regardless, there is a better way to make this required: https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired

main.go
```diff
@@ -140,7 +154,8 @@ If the passed version does not exist, an error is signaled.`)
 - "autoalias": Apply auto aliasing to the provider.`)
 
 	cmd.PersistentFlags().StringVar(&context.UpstreamProviderName, "upstream-provider-name", "",
-		"The name of the upstream provider.")
+		`The name of the upstream provider.
+Must be set in '.upgrade-config.yml' or passed with flag.`)
```
iwahbe, line 158: `MarkFlagRequired` already adds documentation saying it is required. "You can set a flag, `.upgrade-config.yaml` or pass via a flag" is true for all options. It makes sense to mention that part in the general help description, but not on an individual option.
```suggestion
		`The name of the upstream provider.`)
```

upgrade/steps.go
iwahbe: What do we do if `repo.name` is `my-great-provider`?

If this is invalid, lets make sure that we assert that early in the process.

main.go
aq17: I tried `MarkFlagRequired` - it doesn't seem to work when you set `PersistentPreRunE` (https://stackoverflow.com/questions/38105859/make-a-cobra-command-flag-required)

upgrade/upgrade-provider.go
```diff
@@ -57,9 +57,9 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 		discoverSteps = append(discoverSteps,
 			step.F("Planning Provider Update", func() (string, error) {
 				var msg string
-				upgradeTargets, msg, err = GetExpectedTarget(ctx, repoName)
+				upgradeTargets, msg, err = GetExpectedTarget(ctx, repoOrg+"/"+repoName)
 				if err != nil {
-					return "", err
+					return "failed to get expected upgrade version", err
```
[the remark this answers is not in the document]
aq17: No, that was a debugging change.

upgrade/steps.go
aq17: Good point on the assertion. Looking at our [pulumiverse](https://github.com/orgs/pulumiverse/repositories?type=all) providers, i think it's a fair assumption (requirement?) that provider names should start with `pulumi-`.

main.go
iwahbe: It works fine. Validation runs after `PersistentPreRunE`. If that is how `cobra` thinks about validation, we should probably move such validation into our `Run` func. Sorry for the back and forth.

iwahbe (approved):

PR #74, opened 2023-05-05
Allow target upstream version to be passed in

aq17:
Provide a flag `--target-version` to specify the upstream version to upgrade to

### DIFF
--- a/main.go
+++ b/main.go
@@ -161,7 +161,7 @@ If the passed version does not exist, an error is signaled.`)
 		`The kind of upgrade to perform:
 - "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code".
 - "bridge":  Upgrade the bridge only.
-- "provider: Upgrade the upstre am provider only.
+- "provider: Upgrade the upstream provider only.
 - "code":     Perform some number of code migrations.`)
 
 	cmd.PersistentFlags().BoolVar(&experimental, "experimental", false,
--- a/main.go
+++ b/main.go
@@ -144,7 +144,8 @@ If the passed version does not exist, an error is signaled.`)
 		`Use our GH issues to infer the target upgrade version.
 		If both '--target-version' and '--pulumi-infer-version' are passed,
 		we take '--target-version' to cap the inferred version. [Hidden behind PULUMI_DEV]`)
-	cmd.PersistentFlags().MarkHidden("pulumi-infer-version")
+	err := cmd.PersistentFlags().MarkHidden("pulumi-infer-version")
+	contract.AssertNoErrorf(err, "error: %w", err)
 
 	cmd.PersistentFlags().BoolVar(&context.MajorVersionBump, "major", false,
 		`Upgrade the provider to a new major version.`)
--- a/main.go
+++ b/main.go
@@ -144,6 +144,7 @@ If the passed version does not exist, an error is signaled.`)
 		`Use our GH issues to infer the target upgrade version.
 		If both '--target-version' and '--pulumi-infer-version' are passed,
 		we take '--target-version' to cap the inferred version. [Hidden behind PULUMI_DEV]`)
+	cmd.PersistentFlags().MarkHidden("pulumi-infer-version")
 
 	cmd.PersistentFlags().BoolVar(&context.MajorVersionBump, "major", false,
 		`Upgrade the provider to a new major version.`)
--- a/main.go
+++ b/main.go
@@ -27,7 +27,7 @@ const (
 )
 
 func cmd() *cobra.Command {
-	var maxVersion string
+	var inferVersion bool
 	var targetVersion string
 	gopath, ok := os.LookupEnv("GOPATH")
 	if !ok {
@@ -86,15 +86,6 @@ func cmd() *cobra.Command {
 				}
 			}
 
-			// Validate that maxVersion is a valid version
-			if maxVersion != "" {
-				context.MaxVersion, err = semver.NewVersion(maxVersion)
-				if err != nil {
-					return fmt.Errorf("--provider-version=%s: %w",
-						maxVersion, err)
-				}
-			}
-
 			// Validate the kind switch
 			var warnedAll bool
 			for _, kind := range upgradeKind {
@@ -132,7 +123,7 @@ func cmd() *cobra.Command {
 				}
 			}
 
-			if context.MaxVersion != nil && !context.UpgradeProviderVersion {
+			if context.TargetVersion != nil && !context.UpgradeProviderVersion {
 				return fmt.Errorf(
 					"cannot specify the provider version unless the provider will be upgraded")
 			}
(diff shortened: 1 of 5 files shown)


aq17:
Example PR https://github.com/pulumi/pulumi-confluentcloud/pull/134

main.go
```diff
@@ -134,8 +144,13 @@ func cmd() *cobra.Command {
 		},
 	}
 
-	cmd.PersistentFlags().StringVar(&maxVersion, "provider-version", "",
```
aq17, line 137: What was previously `provider-version` is now `max-version`, and the old description for `provider-version` more accurately describes the new `target-version`

iwahbe (commented):
Unless I'm missing something, won't every user who doesn't use our Zapier task pass `--target-version` every time?

What happens if `--max-version` and `--target-version` are both passed?

Since `--max-version` must be a GH issue requesting an upgrade with a corresponding issue, this  feels overly complicated, since `--max-version=${VERSION}` <=> `--target-version=${VERSION}` when both are passed.

We don't expect most of our users to use the GH issue system that we use. What about dropping `--max-version` all together and instead having:

```
--target-version: The version of the upstream provider to upgrade to.

--pulumi-infer-version: Use our GH issues to infer the target upgrade version. If both `--target-version` and `--pulumi-infer-version` are passed, we take `--target-version` to cap the inferred version. [Hidden behind PULUMI_DEV]
```

This way, all options make sense to our end users. We would then add an override in each `.upgrade-config.yaml` setting `--pulumi-infer-version=true`?

iwahbe:
Also, this is wrong:
Screenshot 2023-05-05 at 1 32 19 PM

It does not fix `#0`.

aq17:
> Also, this is wrong: Screenshot 2023-05-05 at 1 32 19 PM
> 
> It does not fix `#0`.
> 
> P.S. We shouldn't need to pass `--upstream-provider-name` since we have https://github.com/pulumi/pulumi-confluentcloud/blob/3f35fb1e8cf0044b5211081feb0f3fcc84b2d205/.upgrade-config.yml#L4.

The config file is detected when running `upgrade-provider` within a provider directory, which is the case when triggered via the github action. Here I just ran it from the `upgrade-provider` directory as I was testing.

aq17:
> Unless I'm missing something, won't every user who doesn't use our Zapier task pass `--target-version` every time?
> 
> What happens if `--max-version` and `--target-version` are both passed?
> 
> Since `--max-version` must be a GH issue requesting an upgrade with a corresponding issue, this feels overly complicated, since `--max-version=${VERSION}` <=> `--target-version=${VERSION}` when both are passed.
> 
> We don't expect most of our users to use the GH issue system that we use. What about dropping `--max-version` all together and instead having:
> 
> ```
> --target-version: The version of the upstream provider to upgrade to.
> 
> --pulumi-infer-version: Use our GH issues to infer the target upgrade version. If both `--target-version` and `--pulumi-infer-version` are passed, we take `--target-version` to cap the inferred version. [Hidden behind PULUMI_DEV]
> ```
> 
> This way, all options make sense to our end users. We would then add an override in each `.upgrade-config.yaml` setting `--pulumi-infer-version=true`?


After this PR, we should add the ability to directly fetch the latest upstream version rather than relying on opened issues, which should  eliminatesthe need for `--target-version`. `--target-version` is provided as an option, but should not be _required_ for users without a similar zapier integration.

Dropping `--max-version` in favor of `--pulumi-infer-version` makes sense

upgrade/upgrade-provider.go
```diff
@@ -64,6 +64,7 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 				}
 
 				// If we have upgrades to perform, we list the new version we will target
+				// unless a version was passed in via `--provider-version`
```
guineveresaenger, line 67: Should this be `--target-version`?

guineveresaenger (approved):
Would like to hold on @iwahbe but this LGTM.

main.go
```diff
@@ -134,11 +135,16 @@ func cmd() *cobra.Command {
...
-		`Upgrade the provider to the passed in version.
+	cmd.PersistentFlags().StringVar(&targetVersion, "target-version", "",
+		`Upgrade the provider to the passed version.
 
 If the passed version does not exist, an error is signaled.`)
 
+	cmd.PersistentFlags().BoolVar(&inferVersion, "pulumi-infer-version", false,
+		`Use our GH issues to infer the target upgrade version.
+		If both '--target-version' and '--pulumi-infer-version' are passed,
+		we take '--target-version' to cap the inferred version. [Hidden behind PULUMI_DEV]`)
```
iwahbe, line 146: This should be actually hidden behind `PULUMI_DEV`. See `cmd.PersistentFlags().MarkHidden`.

upgrade/steps.go
```diff
@@ -494,6 +494,10 @@ func EnsureBranchCheckedOut(ctx Context, branchName string) step.Step {
 //
 // The second argument represents a message to describe the result. It may be empty.
 func GetExpectedTarget(ctx Context, name string) ([]UpgradeTargetIssue, string, error) {
+	if ctx.TargetVersion != nil {
+		return []UpgradeTargetIssue{{Version: ctx.TargetVersion, Number: 0}}, "", nil
```
iwahbe, line 498: This function returns a list of issues that the upgrade will close: `[]UpgradeTargetIssue`. Since the list of issues we are closing is no longer directly tied to the upstream upgrade target like it used to be, can we change the function signature so it accurately represents the data we are returning.

Maybe something like:
```go
type UpstreamUpgradeTarget struct {
    // The version we are targeting. `nil` indicates that no upstream upgrade was found.
    Version *semver.Version
    // The list of issues that this upgrade will close.
    GHIssues []UpgradeTargetIssue
}
```

iwahbe (changes requested):

upgrade/steps.go
```diff
@@ -493,9 +493,11 @@ func EnsureBranchCheckedOut(ctx Context, branchName string) step.Step {
 // sorted by semantic version. The list may be empty.
 //
 // The second argument represents a message to describe the result. It may be empty.
-func GetExpectedTarget(ctx Context, name string) ([]UpgradeTargetIssue, string, error) {
+func GetExpectedTarget(ctx Context, name string) (*UpstreamUpgradeTarget, string, error) {
+	target := &UpstreamUpgradeTarget{Version: ctx.TargetVersion}
 	if ctx.TargetVersion != nil {
-		return []UpgradeTargetIssue{{Version: ctx.TargetVersion, Number: 0}}, "", nil
+		target.Version = ctx.TargetVersion
```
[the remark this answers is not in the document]
aq17: :/

upgrade/steps-helpers.go
```diff
@@ -82,7 +82,7 @@ func originalGoVersionOf(ctx context.Context, repo ProviderRepo, file, needleMod
 	return module.Version{}, false, nil
 }
 
-func prBody(ctx Context, repo ProviderRepo, upgradeTargets UpstreamVersions, goMod *GoMod, targetBridge string) string {
+func prBody(ctx Context, repo ProviderRepo, upgradeTarget *UpstreamUpgradeTarget, goMod *GoMod, targetBridge string) string {
```
iwahbe, line 85: Will `updateTarget` ever be `nil`? If not, this shouldn't be a pointer. If so, then line 109 will panic at
```go
	if len(upgradeTarget.GHIssues) > 0 {
```
If this needs to be a pointer for dataflow reasons, can we add an `contract.Assertf(upgradeTarget != nil`, "upgradeTarget should always be non-nil")`?

iwahbe (approved):
Is `UpstreamVersions` ever used? If not, let's delete it.

LGTM.

PR #76, opened 2023-05-08
Add option to only upgrade `pu/sdk` + `pu/pkg`

aq17:
We should have an option to update only `pu/sdk` and `pu/pu`. This will make it easier to review PR changes and allow us to keep providers up to date more often

### DIFF
--- a/main.go
+++ b/main.go
@@ -102,6 +102,7 @@ func cmd() *cobra.Command {
 				case "all":
 					context.UpgradeBridgeVersion = true
 					context.UpgradeProviderVersion = true
+					context.UpgradeSdkVersion = true
 					if experimental {
 						context.UpgradeCodeMigration = true
 					}
@@ -115,6 +116,8 @@ func cmd() *cobra.Command {
 					set(&context.UpgradeProviderVersion)
 				case "code":
 					set(&context.UpgradeCodeMigration)
+				case "sdk":
+					set(&context.UpgradeSdkVersion)
 				default:
 					return fmt.Errorf(
 						"--kind=%s invalid. Must be one of `all`, `bridge`, `provider`, or `code`.",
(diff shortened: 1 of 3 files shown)


guineveresaenger:
Let's have a summary of what the use case here is. I presume it's when we need to apply an internal fix manually?

upgrade/upgrade-provider.go
```diff
@@ -220,6 +220,15 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 			"go", "get", "github.com/pulumi/pulumi-terraform-bridge/v3@"+targetBridgeVersion)).
 			In(repo.providerDir()))
 	}
+	if ctx.UpgradeSdkVersion {
+		steps = append(steps,
```
iwahbe, line 224: I think its nice if users see a clear correspondence between their inputs and outputs:
```suggestion
		steps = append(steps, step.Combined("Upgrade Pulumi SDK",
```

main.go
```diff
@@ -146,7 +149,8 @@ If the passed version does not exist, an error is signaled.`)
 		`The kind of upgrade to perform:
 - "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code".
```
guineveresaenger, line 150: Should we say `all` includes `sdk` as well?

iwahbe (commented):
We need to update this too:
https://github.com/pulumi/upgrade-provider/blob/8cd32a3f3f84701d7bd472959be9e5abc7ee0380/upgrade/upgrade-provider.go#L150-L155

aq17:
> Let's have a summary of what the use case here is. I presume it's when we need to apply an internal fix manually?

I believe @t0yv0 suggested this could be beneficial for reviewing PRs especially

t0yv0:
Yes updating `github.com/pulumi/pulumi/pkg/v3` ref can bring in changes from codegen - and generate diffs. These changes happen on a weekly cadence released together with pulumi/pulumi. I thought it'd be marvelous to not couple them with updates initiated by new upstream versions or new bridge versions or some such.

guineveresaenger:
brilliant, thanks everyone 🙌

PR #81, opened 2023-05-11
Set infer version flag value in Context

aq17:
`--pulumi-infer-version` was being unused

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,6 +62,8 @@ func cmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			context.InferVersion = inferVersion
+
 			// Validate argument is {org}/{repo}
 			tok := strings.Split(args[0], "/")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -62,6 +62,8 @@ func cmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			context.InferVersion = inferVersion
```
iwahbe, line 65: Why do we need `inferVersion`? Can we not reference `context.InferVersion` directly?

iwahbe (approved):
This looks right to me. Can you run a test upgrade and confirm this works as expected before merging?

aq17:
> This looks right to me. Can you run a test upgrade and confirm this works as expected before merging?

https://github.com/pulumi/pulumi-fastly/pull/263 ✔️

PR #83, opened 2023-05-12
Allow expected repo path to be set in config

### LINKED ISSUE #53: Allow expected repo path to be set in config
Part of https://github.com/pulumi/upgrade-provider/issues/52
The expected repo path should be able to be set as a config variable, i.e. `REPO_PATH`. The order of precedence should be:
1) `REPO_PATH` is set
2) `upgrade-provider` is being called from within a provider directory/ sub-directory (see https://github.com/pulumi/upgrade-provider/pull/46)
3) Default set to $GOPATH (https://github.com/pulumi/upgrade-provider/blob/main/main.go#L19-L27)

aq17:
Allow the user to specify the desired path to clone the provider repo.

Fixes #53


upgrade/util.go
```diff
@@ -12,7 +12,8 @@ import (
 type Context struct {
 	context.Context
 
-	GoPath string
+	GoPath   string
+	RepoPath string
```
iwahbe, line 16: Lets add some documentation explaining what this means.

iwahbe (commented):
Is there a way we can architect the data-flow so that `ctx.RepoPath` isn't always available but sometime accurate?

aq17:
> Is there a way we can architect the data-flow so that `ctx.RepoPath` isn't always available but sometime accurate?

Not fully sure what you mean – are you suggesting not having `ctx.RepoPath` if empty? We only use `ctx.GoPath` to get the expected location, so we could just have one variable `RepoPath` that is set to $GOPATH by default, and can be overwritten with `--repo-path`?

iwahbe (approved):

upgrade/util.go
```diff
@@ -31,6 +34,10 @@ type Context struct {
 	RemovePlugins bool
 }
 
+func (c Context) SetRepoPath(p string) {
+	c.repoPath = &p
```
iwahbe, line 38:
```suggestion
	*c.repoPath = p
```

PR #98, opened 2023-06-14
Create issue on upgrade provider failure

### LINKED ISSUE #97: Alert maintainers on failure to run upgrade-provider
When the upgrade-provider tool fails to run, we should alert maintainers that automatic upgrades may be failing.

Creating a GitHub Issue would be appropriate, if we could ensure that we only update/replace a particular open issue by name to avoid spamming a repository.

### Background

Pulumi Azure has had a series of [upgrade provider failures](https://github.com/pulumi/pulumi-azure/actions/workflows/upgrade-provider.yml), some stemming from a [dependency issue in the bridge](https://github.com/pulumi/pulumi-azure/actions/runs/5252151683/jobs/9487857739) or due to errors following attempted automatic token mapping:

```
Falling back to direct application...
panic: fatal: error An assertion has failed: failed to compute default tokens. source error: 2 errors occurred:
        * resources:
3 errors occurred:
        * could not find a module that prefixes 'arc_machine_extension'
```

aq17:
Resolves #97 

This PR creates an issue in the targeted provider repo when `upgrade-provider` fails. The title is `Upgrade provider failure: {year} {month} {date}`. 
It looks for existing issues of the form `Upgrade provider failure: ` first, and deletes them if existing, since we likely only need to be aware of the latest upgrade failure. 
The issue body is the error message emitted from the `upgrade-provider` attempt.
It adds the "needs-triage" label (teams are not allowed to be passed as the `--assignee`, so couldn't use `pulumi/Ecosystem` for example)

### DIFF
--- a/main.go
+++ b/main.go
@@ -274,8 +274,8 @@ func createFailureIssue(ctx upgrade.Context, repoOrg string, repoName string) (s
 		"--limit=1",
 		"--json=url",
 	)
-	bytes1 := new(bytes.Buffer)
-	getLatestWorkflowRun.Stdout = bytes1
+	ghRunBytes := new(bytes.Buffer)
+	getLatestWorkflowRun.Stdout = ghRunBytes
 	err := getLatestWorkflowRun.Run()
 	if err != nil {
 		return "createFailureIssue error: failed to query latest workflow run", err
@@ -283,7 +283,7 @@ func createFailureIssue(ctx upgrade.Context, repoOrg string, repoName string) (s
 	var workflowRunUrls []struct {
 		Url string `json:"url"`
 	}
-	err = json.Unmarshal(bytes1.Bytes(), &workflowRunUrls)
+	err = json.Unmarshal(ghRunBytes.Bytes(), &workflowRunUrls)
 	if err != nil {
 		return "createFailureIssue error: failed to unmarshal `gh run list` output", err
 	}
@@ -299,7 +299,8 @@ func createFailureIssue(ctx upgrade.Context, repoOrg string, repoName string) (s
 		"--state=open",
 		"--author=pulumi-bot",
 	)
-	getIssues.Stdout = new(bytes.Buffer)
+	ghSearchBytes := new(bytes.Buffer)
+	getIssues.Stdout = ghSearchBytes
 	err = getIssues.Run()
 	if err != nil {
 		return "createFailureIssue error: failed to search existing issues", err
@@ -308,7 +309,7 @@ func createFailureIssue(ctx upgrade.Context, repoOrg string, repoName string) (s
 		Title  string `json:"title"`
 		Number int    `json:"number"`
 	}{}
-	err = json.Unmarshal(getIssues.Stdout.Bytes(), &titles)
+	err = json.Unmarshal(ghSearchBytes.Bytes(), &titles)
 	if err != nil {
 		return "createFailureIssue error: failed to unmarshal `gh search issues` output", err
 	}
--- a/main.go
+++ b/main.go
@@ -142,7 +142,7 @@ func cmd() *cobra.Command {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
 			if err != nil {
 				msg, err := createFailureIssue(context, repoOrg, repoName)
-				if err != nil {
+				if err != nil && context.InferVersion {
 					fmt.Println(msg)
 				}
 			}
--- a/main.go
+++ b/main.go
@@ -140,9 +140,9 @@ func cmd() *cobra.Command {
 		},
 		Run: func(_ *cobra.Command, args []string) {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
-			if err != nil {
+			if err != nil && context.InferVersion {
 				msg, err := createFailureIssue(context, repoOrg, repoName)
-				if err != nil && context.InferVersion {
+				if err != nil {
 					fmt.Println(msg)
 				}
 			}
--- a/main.go
+++ b/main.go
@@ -141,7 +141,7 @@ func cmd() *cobra.Command {
 		Run: func(_ *cobra.Command, args []string) {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
 			if err != nil {
-				msg, err := createFailureIssue(context, repoOrg, repoName, err.Error())
+				msg, err := createFailureIssue(context, repoOrg, repoName)
 				if err != nil {
 					fmt.Println(msg)
 				}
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -246,3 +256,52 @@ func bindFlags(cmd *cobra.Command, v *viper.Viper) {
...
+
+// Create an issue in the provider repo with a message describing the upgrade failure
+func createFailureIssue(ctx upgrade.Context, repoOrg string, repoName string, errMsg string) (string, error) {
+	now := time.Now()
+	y, m, d := now.Date()
+	title := fmt.Sprintf("Upgrade provider failure: %v %v %v", y, m, d)
+
+	getIssues := exec.CommandContext(ctx, "gh", "search", "issues",
+		"Upgrade provider failure: ",
+		"--repo="+repoOrg+"/"+repoName,
```
AaronFriel, line 268: Can or should we also search for author here?

main.go
```diff
@@ -246,3 +256,52 @@ func bindFlags(cmd *cobra.Command, v *viper.Viper) {
...
+		Title  string `json:"title"`
+		Number int    `json:"number"`
+	}{}
+	err = json.Unmarshal(bytes.Bytes(), &titles)
+	if err != nil {
+		return "Failed to create failure issue: failed to unmarshal gh output", err
+	}
+	for _, t := range titles {
+		cmd := exec.Command("gh", "issue", "delete",
+			fmt.Sprintf("%v", t.Number),
```
AaronFriel, line 287: I don't think we should delete here; as I think we expect issues to persist.

Comment on the issue instead that a new run failed with a link to the workflow ID?

main.go
```diff
@@ -136,6 +140,12 @@ func cmd() *cobra.Command {
 		},
 		Run: func(_ *cobra.Command, args []string) {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
+			if err != nil {
+				msg, err := createFailureIssue(context, repoOrg, repoName, err.Error())
```
AaronFriel, line 144: Probably want to format this error message to include a link to the workflow that failed.

I think we can access that via a GitHub env var.

main.go
```diff
@@ -136,6 +140,12 @@ func cmd() *cobra.Command {
 		},
 		Run: func(_ *cobra.Command, args []string) {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
+			if err != nil {
```
AaronFriel, line 143: If not running in GitHub CI, skip the issue creation?

iwahbe:
The linked example has been deleted. I'd like to see the example, and I suspect we will want additional formatting beyond the error message itself. We should link to the failed workflow run if run in CI. 

Most importantly, `upgrade.UpgradeProvider` was designed to handle displaying errors to the user. If an error is signaled in one of the steps, `step.Run` displays the error to the user and returns `false`. `upgrade.UpgradeProvider` returns this:

https://github.com/pulumi/upgrade-provider/blob/38ae9cf189e17f53637a87c1da4cb6c725a1a8e1/upgrade/util.go#L42-L49

We can change this paradigm to make the returned error meaningful, but we need to do that.

iwahbe (commented):
Instead of deleting old issues, can we edit the existing issue with the new data.

Also, see my other comment above.

main.go
aq17: We can assume `pulumi-bot` so we are updated on automation failures – will add this later after testing

main.go
aq17: We can reuse the `pulumi-infer-version` flag for this

aq17:
Updated issue example: https://github.com/pulumi/pulumi-akamai/issues/214

@iwahbe I think that for now, linking to the workflow run along with a timestamp might be sufficient, instead of refactoring the step/ error paradigm. Is that okay with you?

iwahbe:
> Updated issue example: [pulumi/pulumi-akamai#214](https://github.com/pulumi/pulumi-akamai/issues/214) The initial issue and first two comments have my username as I ran it from the CLI, so the latest workflow is irrelevant; the last comment from `pulumi-bot` was created by triggering `upgrade-provider` in CI, which correctly links to that workflow
> 
> @iwahbe I think that for now, linking to the workflow run along with a timestamp might be sufficient, instead of refactoring the step/ error paradigm. Is that okay with you?

That sounds good to me.

main.go
```diff
@@ -246,3 +256,87 @@ func bindFlags(cmd *cobra.Command, v *viper.Viper) {
...
+	}
+	if len(workflowRunUrls) == 0 {
+		return "createFailureIssue error: no workflow runs detected", err
+	}
+	workflowUrl := workflowRunUrls[0].Url
+
+	getIssues := exec.CommandContext(ctx, "gh", "search", "issues",
+		"Upgrade provider failure: ",
+		"--repo="+repoOrg+"/"+repoName,
+		"--json=title,number",
```
iwahbe, line 298: We only want open issues here.
```suggestion
		"--json=title,number",
		"--state=open",
```

main.go
```diff
@@ -246,3 +256,87 @@ func bindFlags(cmd *cobra.Command, v *viper.Viper) {
...
+	getIssues.Stdout = bytes2
+	err = getIssues.Run()
+	if err != nil {
+		return "createFailureIssue error: failed to search existing issues", err
+	}
+	titles := []struct {
+		Title  string `json:"title"`
+		Number int    `json:"number"`
+	}{}
+	err = json.Unmarshal(bytes2.Bytes(), &titles)
```
iwahbe, line 310: `bytes2` can just be inlined, and by inlining we can make clear what the purpose of `bytes2` actually is.
```suggestion
	getIssues.Stdout = new(bytes.Buffer)
	err = getIssues.Run()
	if err != nil {
		return "createFailureIssue error: failed to search existing issues", err
	}
	titles := []struct {
		Title  string `json:"title"`
		Number int    `json:"number"`
	}{}
	err = json.Unmarshal(getIssues.Stdout.Bytes(), &titles)
```

main.go
```diff
@@ -246,3 +256,87 @@ func bindFlags(cmd *cobra.Command, v *viper.Viper) {
...
+	if err != nil {
+		return "createFailureIssue error: failed to unmarshal `gh search issues` output", err
+	}
+
+	// create new issue if none exist
+	if len(titles) == 0 {
+		cmd := exec.Command(
+			"gh", "issue", "create",
+			"--repo="+repoOrg+"/"+repoName,
+			"--body="+fmt.Sprintf("View the latest workflow run: %s", workflowUrl),
```
iwahbe, line 320: Can we keep the `fmt.Sprintf` style consistent?
```suggestion
			"--repo="+repoOrg+"/"+repoName,
			"--body=View the latest workflow run: "+workflowUrl,
```

aq17:
@iwahbe mind taking a quick look on this when you can?

main.go
```diff
@@ -136,6 +140,12 @@ func cmd() *cobra.Command {
 		},
 		Run: func(_ *cobra.Command, args []string) {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
+			if err != nil {
+				msg, err := createFailureIssue(context, repoOrg, repoName)
+				if err != nil && context.InferVersion {
+					fmt.Println(msg)
+				}
```
iwahbe, line 147: This doesn't make any sense to me. If we try to create the issue and fail, we should always report to the user. 

If we're gating issue creation on a Pulumi provider, the `&& context.InferVersion` check should be on line 143, before we try to create an issue.

iwahbe (approved):
One piece of logic doesn't seem right. The rest looks good.

main.go
aq17: You are correct

PR #99, opened 2023-06-14
Remove sdk upgrade step from --all flag

### LINKED ISSUE #96: Upgrading pulumi/pulumi:`latest` results in failing upgrades when bridge dependencies are on older version
### What happened?

When we upgrade `kind:all`, the upgrade-provider tool attempts to upgrade to the latest released version of pulumi/pulumi,

This can result in failures on the event that the terraform bridge is using an older version of pulumi/pulumi.

### Expected Behavior

The provider should build successfully in CI.

### Steps to reproduce

Sample upgrade failures:

- https://github.com/pulumi/pulumi-cloudamqp/commit/3c7019ffb0eddadd34120744ed5ec35ea11b60c3/checks
- https://github.com/pulumi/pulumi-civo/commit/7a805297c1d0871d062ea9501f05cd6453ec0760/checks

### Output of `pulumi about`

n/a

### Additional context

A quick fix for this situation is to upgrade the bridge to using p/p 3.70 but a longer term fix is to tweak `kind:all` to use the bridge dependency version for pulumi/pulumi.

### Contributing

guineveresaenger:
Fixes #96.

A successful upgrade PR using these changes can be seen [on pulumi-civo](https://github.com/pulumi/pulumi-civo/pull/242/files#diff-0301265a2d4321ca487bbd7ba523f69633171cef680c0114e7ca8d2c34176b73R10)
As you can see, we did not upgrade the Pulumi version explicitly and yet go mod tidy pulled in the version the bridge is currently using, 3.71.0.

I did discover an issue with the new pr-reviewers flag; issue to follow.

### DIFF
--- a/main.go
+++ b/main.go
@@ -103,13 +103,20 @@ func cmd() *cobra.Command {
 				case "all":
 					context.UpgradeBridgeVersion = true
 					context.UpgradeProviderVersion = true
-					context.UpgradeSdkVersion = true
 					if experimental {
 						context.UpgradeCodeMigration = true
 					}
+
 					if len(upgradeKind) > 1 {
-						warnedAll = true
-						warn("`--kind=all` implies all other options")
+						for _, v := range upgradeKind {
+							if v == "sdk" {
+								context.UpgradeSdkVersion = true
+							}
+						}
+						if !context.UpgradeSdkVersion {
+							warnedAll = true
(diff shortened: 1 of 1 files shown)


aq17 (approved):

iwahbe:
It seems a bit confusing that `--kind=all` no longer implies that all kinds of upgrades are happening. I think we have some naming issues.

PR #107, opened 2023-07-03
Only create an issue when run in CI

iwahbe:
Running CI locally shouldn't create an issue in GH.

### DIFF
--- a/main.go
+++ b/main.go
@@ -148,9 +148,16 @@ func cmd() *cobra.Command {
 		Run: func(_ *cobra.Command, args []string) {
 			err := upgrade.UpgradeProvider(context, repoOrg, repoName)
 			if err != nil && context.InferVersion {
-				msg, err := createFailureIssue(context, repoOrg, repoName)
-				if err != nil {
-					fmt.Println(msg)
+				// $GITHUB_ACTION is a default env var within github
+				// actions, but is unlikely to be defined elsewhere.
+				//
+				// We do this to test that we are being run inside a GH
+				// action.
+				if _, ci := os.LookupEnv("GITHUB_ACTION"); ci {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -275,40 +275,26 @@ func createFailureIssue(ctx upgrade.Context, repoOrg string, repoName string) (s
...
+		os.Getenv("GITHUB_REPOSITORY"),
+		os.Getenv("GITHUB_RUN_ID"))
 
 	getIssues := exec.CommandContext(ctx, "gh", "search", "issues",
 		"Workflow failure: Upgrade provider",
 		"--repo="+repoOrg+"/"+repoName,
 		"--json=title,number",
 		"--state=open",
-		"--author=pulumi-bot",
+		"--author=@me",
```
iwahbe, line 293: When CI is run locally, it always creates a new issue. This is categorically fixed by not hard-coding `pulumi-bot` and instead referencing the current user.

This will also improve the experience for any user who isn't Pulumi.

main.go
```diff
@@ -148,9 +148,16 @@ func cmd() *cobra.Command {
...
 			if err != nil && context.InferVersion {
-				msg, err := createFailureIssue(context, repoOrg, repoName)
-				if err != nil {
-					fmt.Println(msg)
+				// $GITHUB_ACTION is a default env var within github
+				// actions, but is unlikely to be defined elsewhere.
+				//
+				// We do this to test that we are being run inside a GH
+				// action.
+				if _, ci := os.LookupEnv("GITHUB_ACTION"); ci {
```
iwahbe, line 156: We only want to run this in Github actions CI. Otherwise the link we set doesn't make sense.

aq17 (approved):

PR #116, opened 2023-07-21
Fix argument order to `gitRefsOf`

aq17:
Order of arguments just needs to be flipped, was causing failures such as https://github.com/pulumi/pulumi-artifactory/actions/runs/5625475142/job/15244327189


aq17:
On an unrelated note, when I try to rerun the above upgrade with this change, I still get a 403 `error: [resource plugin time-0.0.15] downloading from : failed to download plugin: time-0.0.15: 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-time-v0.0.15-darwin-arm64.tar.gz` even with a sufficient access token

iwahbe (approved):
We really need to spend the time to get a test framework here. Thanks for the fix!

PR #118, opened 2023-07-26
Initialize submodule for patched providers before modifying go.mod

### LINKED ISSUE #112: CI upgrades not initializing upstream repo correctly
The last week or so, most upgrades have been failing with a similar issue
https://github.com/pulumi/pulumi-gitlab/commit/74bb1cc35db6178e918cf9a1b5b3f94739025d99/checks

`go: gitlab.com/gitlab-org/terraform-provider-gitlab@v0.0.0 (replaced by ../upstream): reading ../upstream/go.mod: open /home/runner/work/pulumi-gitlab/pulumi-gitlab/upstream/go.mod: no such file or directory`

where it appears the upstream directory is not being initialized/ is empty, causing the upgrade to fail.

Re-running the upgrade locally succeeds (i.e. https://github.com/pulumi/pulumi-gitlab/pull/316)

aq17:
Fixes #112 

Some updates in CI were failing for patched providers only when the TF plugin SDK needed to be updated (https://github.com/pulumi/pulumi-azure/actions/runs/5617365095/job/15221349196#step:3:1246), because the upstream submodule was not initialized first.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -169,23 +169,7 @@ func UpgradeProviderVersion(
 	ctx Context, goMod *GoMod, target *semver.Version,
 	repo ProviderRepo, targetSHA, forkedProviderUpstreamCommit string,
 ) step.Step {
-	updateLatestPluginSDK, didUpdate := getLatestTFPluginSDKReplace(ctx, repo)
-
-	// We start by updating the terraform-plugin-sdk because later updates sometimes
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
This looks reasonable. Make sure that we have test against a patched and an unpatched provider before merging.

Otherwise, LGTM!

PR #119, opened 2023-07-26
Add update plugin framework as an option

### LINKED ISSUE #89: Add update plugin framework as an option
## Hello!

## Issue details

We have autoupdates set for `pulumi-terraform-bridge`. We should also include updates to the `pulumi-terraform-bridge/pf` module.

### Affected area/feature

Automating module updates such as [this update to the muxer](https://github.com/pulumi/pulumi-terraform-bridge/pull/1142), which prompted a new version of the `pf` module.

aq17:
Fixes #89 

Add option to update plugin framework (i.e. `upgrade-provider --kind=pf`).

### DIFF
--- a/main.go
+++ b/main.go
@@ -107,6 +107,7 @@ func cmd() *cobra.Command {
 				case "all":
 					context.UpgradeBridgeVersion = true
 					context.UpgradeProviderVersion = true
+					context.UpgradePfVersion = true
 					if experimental {
 						context.UpgradeCodeMigration = true
 					}
@@ -186,7 +187,7 @@ If the passed version does not exist, an error is signaled.`)
 
 	cmd.PersistentFlags().StringSliceVar(&upgradeKind, "kind", []string{"all"},
 		`The kind of upgrade to perform:
-- "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code".
+- "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code,pf".
 - "bridge":  Upgrade the bridge only.
 - "provider": Upgrade the upstream provider only.
 - "sdk": Upgrade the Pulumi sdk only.
--- a/main.go
+++ b/main.go
@@ -130,6 +130,8 @@ func cmd() *cobra.Command {
 					set(&context.UpgradeCodeMigration)
 				case "sdk":
 					set(&context.UpgradeSdkVersion)
+				case "pf":
+					set(&context.UpgradePfVersion)
 				default:
 					return fmt.Errorf(
 						"--kind=%s invalid. Must be one of `all`, `bridge`, `provider`, `code`, or `sdk`.",
@@ -188,6 +190,7 @@ If the passed version does not exist, an error is signaled.`)
 - "bridge":  Upgrade the bridge only.
 - "provider": Upgrade the upstream provider only.
 - "sdk": Upgrade the Pulumi sdk only.
+- "pf": Upgrade the Plugin Framework only.
 - "code":     Perform some number of code migrations.`)
 
 	cmd.PersistentFlags().BoolVar(&experimental, "experimental", false,
--- a/upgrade/kind.go
+++ b/upgrade/kind.go
@@ -99,10 +99,13 @@ func GetRepoKind(ctx Context, repo ProviderRepo) (*GoMod, error) {
 	} else if !ok {
 		return nil, fmt.Errorf(bridgeMissingMsg)
 	}
+
 	pf, ok, err := originalGoVersionOf(ctx, repo, filepath.Join("provider", "go.mod"), "github.com/pulumi/pulumi-terraform-bridge/pf")
-	if (err != nil || !ok) && ctx.UpgradePfVersion {
-		fmt.Println("No version of github.com/pulumi/pulumi-terraform-bridge/pf detected")
-		ctx.UpgradePfVersion = false
+	if err != nil {
+		return nil, err
+	}
+	if !ok && ctx.UpgradeProviderVersion {
+		setFalse(&ctx.UpgradeProviderVersion)
 	}
 
 	tfProviderRepoName := ctx.UpstreamProviderName
@@ -207,4 +210,8 @@ func GetRepoKind(ctx Context, repo ProviderRepo) (*GoMod, error) {
 	}
 
 	return &out, nil
+}
+
+func setFalse(b *bool) {
+	*b = false
 }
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -130,6 +130,8 @@ func cmd() *cobra.Command {
 					set(&context.UpgradeCodeMigration)
 				case "sdk":
 					set(&context.UpgradeSdkVersion)
+				case "pf":
```
guineveresaenger, line 133: Should this be included in the `--all` flag?

aq17: I don't think so because not all providers utilize it?

guineveresaenger: Would this result in confusion if we have a provider that does use it, and we run upgrade --all?I realize this is more about managing expectations.

upgrade/upgrade-provider.go
```diff
@@ -305,6 +305,11 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 				"go", "get", "github.com/pulumi/pulumi/pkg/v3")).
 				In(repo.examplesDir()))
 	}
+	if ctx.UpgradePfVersion {
+		steps = append(steps, step.Cmd(exec.CommandContext(ctx,
+			"go", "get", "github.com/pulumi/pulumi-terraform-bridge/pf")).
+			In(repo.providerDir()))
+	}
```
iwahbe, line 312: This should be similar to upgrading the bridge version (with an explicit target version) and probably occur right afterwards.

iwahbe (changes requested):
I agree with @guineveresaenger. `pf` needs to be part of `--all` so it is included by default. Normal users (us included) should not need to pass `--kind` flags to update everything to latest.

This update should work the same way as updating `github.com/pulumi/pulumi-terraform-bridge/v3`. It needs to:

1. Detect the presence of `github.com/pulumi/pulumi-terraform-bridge/pf` in the planning step and indicate the target version it will seek. 
  a. If no version of `github.com/pulumi/pulumi-terraform-bridge/pf` was detected and `--all` was passed, we just toggle `UpgradePfVersion` off. If `--pf` was passed explicitly, we should error here.
2. Upgrade to the target version.
3. Report in the GH PR body about the upgrade: `old -> new`.


Hopefully we can reuse a lot of code from the `github.com/pulumi/pulumi-terraform-bridge/v3` update.

upgrade/upgrade-provider.go
```diff
@@ -187,6 +187,24 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 			}).AssignTo(&tfSDKUpgrade),
 		)
 	}
+	if ctx.UpgradePfVersion {
+		discoverSteps = append(discoverSteps,
+			step.F("Planning Plugin Framework Update", func() (string, error) {
+				latest, err := latestRelease(ctx, "pulumi/pulumi-terraform-bridge")
```
aq17, line 193: This part needs to be fixed to detect the latest `pf/*` tag

iwahbe: This will need to look at tags, since we don't "release" `pf/vX.Y.Z` according to GH.

upgrade/kind.go
```diff
@@ -99,6 +99,11 @@ func GetRepoKind(ctx Context, repo ProviderRepo) (*GoMod, error) {
 	} else if !ok {
 		return nil, fmt.Errorf(bridgeMissingMsg)
 	}
+	pf, ok, err := originalGoVersionOf(ctx, repo, filepath.Join("provider", "go.mod"), "github.com/pulumi/pulumi-terraform-bridge/pf")
+	if (err != nil || !ok) && ctx.UpgradePfVersion {
```
iwahbe, line 103: `!ok` is different then an `err != nil`. If we get an error, we should report it and abort. `!ok` just means it wasn't found. This is expected behavior for lots of our providers.

upgrade/upgrade-provider.go
aq17: I might be missing an obvious solution, but I can't seem to find a simple command to list, or filter, tags by name using `gh`.

iwahbe: I believe that there is one, but I've been using `git ls-remote` for this:

https://github.com/pulumi/upgrade-provider/blob/cb42bff41add5feada028094f5fc02922fbdbd0b/upgrade/steps-helpers.go#L308-L309

upgrade/upgrade-provider.go
```diff
@@ -190,19 +190,30 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 	if ctx.UpgradePfVersion {
 		discoverSteps = append(discoverSteps,
 			step.F("Planning Plugin Framework Update", func() (string, error) {
-				latest, err := latestRelease(ctx, "pulumi/pulumi-terraform-bridge")
+				refs, err := gitRefsOf(ctx, "https://"+modPathWithoutVersion(goMod.Bridge.Path),
+					"tags")
 				if err != nil {
 					return "", err
 				}
+				sortedRefs := refs.sortedLabels(func(a string, b string) bool { return a > b })
```
aq17, line 198: I'm waiting for a pf based provider upgrade to test this with

guineveresaenger: hi! you can test this with pulumi-cloudflare! there's a [patch upgrade of bridge/pf](https://github.com/pulumi/pulumi-terraform-bridge/releases/tag/pf%2Fv0.15.1) and cloudflare could use that patched through.

aq17: I think it looks good now https://github.com/pulumi/pulumi-cloudflare/pull/462

aq17: @guineveresaenger do you mind giving this another pass when possible?

upgrade/upgrade-provider.go
```diff
@@ -173,6 +174,35 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
...
+				for _, r := range sortedRefs {
+					if v := strings.TrimPrefix(r, "refs/tags/pf/"); v != r {
+						if targetVersion.LessThan(semver.MustParse(v)) {
+							targetVersion = *semver.MustParse(v)
+						}
+					}
+				}
+
+				// If our target upgrade version is the same as our current version, we skip the update.
+				if targetVersion.String() == goMod.Pf.Version {
```
iwahbe, line 197: https://github.com/pulumi/upgrade-provider/pull/125/ introduced a function to do this.
```suggestion
				targetVersion := latestSemverTag("pf/", refs)

				// If our target upgrade version is the same as our current version, we skip the update.
				if targetVersion.Original() == goMod.Pf.Version {
```

iwahbe (approved):
One suggestion, otherwise LGTM

Issue #121, opened 2023-07-30
allow customizing the pr assignee


omercnet:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
When running this in a github action the GITHUB_TOKEN belongs to @github-actions[bot], so using `@me` as the assignee fails, allowing to customize the assignee would help mitigate that

rquitales:
@omercnet Thanks for reporting this enhancement and supplying a PR to address this. We look forward to working with you on merging this in.

PR #122, opened 2023-07-30
feat: add a pr assign config var

omercnet:
fixes https://github.com/pulumi/upgrade-provider/issues/121

### DIFF
--- a/main.go
+++ b/main.go
@@ -210,7 +210,7 @@ Required unless running from provider root and set in upgrade-config.yml.`)
 		`A comma separated list of reviewers to assign the upgrade PR to.`)
 
 	cmd.PersistentFlags().StringVar(&context.PrAssign, "pr-assign", "@me",
-		`A user to assign the upgrade PR to (default: @me).`)
+		`A user to assign the upgrade PR to.`)
 
 	cmd.PersistentFlags().BoolVarP(&context.AllowMissingDocs, "allow-missing-docs", "", false,
 		`If true, don't error on missing docs during tfgen.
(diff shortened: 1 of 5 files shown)


rquitales:
The changes here look good to me, but I'm unfamiliar with the codebase, so will defer to other team members for final approval.

main.go
```diff
@@ -209,6 +209,9 @@ Required unless running from provider root and set in upgrade-config.yml.`)
 	cmd.PersistentFlags().StringVar(&context.PrReviewers, "pr-reviewers", "",
 		`A comma separated list of reviewers to assign the upgrade PR to.`)
 
+	cmd.PersistentFlags().StringVar(&context.PrAssign, "pr-assign", "@me",
+		`A user to assign the upgrade PR to (default: @me).`)
```
iwahbe, line 213: The default is already included by cobra, so this renders as 

```
--pr-assign string        A user to assign the upgrade PR to (default: @me). (default "@me")
```

```suggestion
	cmd.PersistentFlags().StringVar(&context.PrAssign, "pr-assign", "@me",
		`A user to assign the upgrade PR to.`)
```

iwahbe (changes requested):
Hi @omercnet. Thanks for the PR!

To finish up, you will need to thread `Context.PrAssign` to where `"@me"` is currently used:

https://github.com/pulumi/upgrade-provider/blob/cb42bff41add5feada028094f5fc02922fbdbd0b/upgrade/steps.go#L313-L315

Right now, your changes give the user the option to set their desired assignee but doesn't actually respect that option.

iwahbe (approved):
This looks good to me.

Issue #123, opened 2023-08-02
Major upgrade to v2 mangles SDK go.mod


danielrbradley:
### What happened?

Running `upgrade-provider pulumi/pulumi-wavefront --major` going from v1 to v2 the go.mod is updated as:

```diff
- module github.com/pulumi/pulumi-wavefront/sdk
+ module github.com/pulumi/pulumi-wavefront/sdkv2
```

### Expected Behavior

```diff
- module github.com/pulumi/pulumi-wavefront/sdk
+ module github.com/pulumi/pulumi-wavefront/sdk/v2
```

### Steps to reproduce

1. Checkout v1 of pulumi-wavefront
2. Run `upgrade-provider pulumi/pulumi-wavefront --major`

### Output of `pulumi about`

N/A

### Additional context

Using version v0.0.0-20230731114225-f9385d3290cb of upgrade-provider.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #129, opened 2023-08-28
Correctly upgrade sdk/go.mod during major version bumps

iwahbe:
This normalizes how we modify `sdk/go.mod`, and in doing so corrects the string we are modifying so it actually works.

Fixes https://github.com/pulumi/upgrade-provider/issues/123

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 
@@ -555,7 +556,7 @@ func MajorVersionBump(ctx Context, goMod *GoMod, target *UpstreamUpgradeTarget, 
 		prev += fmt.Sprintf("/v%d", repo.currentVersion.Major())
 	}
 
-	nextMajorVersion := fmt.Sprintf("v%d", repo.currentVersion.Major()+1)
+	nextMajorVersion := fmt.Sprintf("v%d", repo.currentVersion.IncMajor().Major())
 
 	// Replace s in file, where {} is interpolated into the old and new provider
 	// component of the path.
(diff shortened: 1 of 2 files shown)


upgrade/steps.go
```diff
@@ -574,27 +575,34 @@ func MajorVersionBump(ctx Context, goMod *GoMod, target *UpstreamUpgradeTarget,
...
+		UpdateFile("Update Go Module in provider/go.mod", "go.mod",
+			func(src []byte) ([]byte, error) {
+				base := "module github.com/pulumi/" + name + "/provider"
+				old := base
+				if repo.currentVersion.Major() > 1 {
+					old += fmt.Sprintf("/v%d", repo.currentVersion.Major())
+				}
+				new := base + fmt.Sprintf("/v%d", repo.currentVersion.IncMajor().Major())
+
+				return bytes.ReplaceAll(src, []byte(old), []byte(new)), nil
```
mjeffryes, line 595: I think I'm missing something; how is this different than the 
```
replaceInFile("Update Go Module", "go.mod",
			"module github.com/pulumi/"+name+"/{}").In(repo.providerDir()),
```
step it replaced?

upgrade/steps.go
```diff
@@ -574,27 +575,34 @@ func MajorVersionBump(ctx Context, goMod *GoMod, target *UpstreamUpgradeTarget,
...
+			}).In(repo.providerDir()),
+		UpdateFile("Update Go Module in sdk/go.mod", "go.mod", func(src []byte) ([]byte, error) {
```
mjeffryes, line 597: FWIW, it seems like you could also make the `ReplaceAll` helper just manage the version number (dropping the "provider" prefix) and use it for this step as well.

mjeffryes (approved):
LGTM. Some non-blocking comments

upgrade/steps.go
iwahbe: I exploded it when I was debugging. Its' not. https://github.com/pulumi/upgrade-provider/pull/129/commits/b6a42f75badf46f22f7bbb34e31bac3b31025b31 has compressed it back down.

upgrade/steps.go
iwahbe: https://github.com/pulumi/upgrade-provider/pull/129/commits/b6a42f75badf46f22f7bbb34e31bac3b31025b31 does that. Thanks for the suggestion.

PR #131, opened 2023-08-31
Exit non-zero on error

### LINKED ISSUE #130: does not fail on error
### What happened?

running the action with a malformed config fails the command but the action does not fail

### Expected Behavior

action should fail, which would result in a notification to the repo admin that something's broken

### Steps to reproduce

see https://github.com/pulumiverse/pulumi-vercel/pull/44 and all the runs in https://github.com/pulumiverse/pulumi-vercel/actions/runs/6001806781/job/16276798111 that were successful although the action failed

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #130

We change how errors are displayed on invalid config files. The error message for an invalid config file now looks like this:

```sh
𝛌 upgrade-provider pulumi/pulumi-aiven
error: While parsing config: yaml: line 4: did not find expected ',' or '}'
```

It used to look like this:

```sh
𝛌 upgrade-provider pulumi/pulumi-aiven
Error: While parsing config: yaml: line 4: did not find expected ',' or '}'
Usage:
  upgrade-provider <provider> [flags]

Flags:
      --allow-missing-docs              If true, don't error on missing docs during tfgen.
                                        This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.
      --create-failure-issue            Create an issue in the target repository if the upgrade attempt fails in CI.
      --experimental                    Enable experimental features, such as auto token mapping and auto aliasing

...
```

---

As a side note, this will cause failed upgrades to show up (correctly) as failed workflows.

### DIFF
--- a/main.go
+++ b/main.go
@@ -57,6 +57,11 @@ func cmd() *cobra.Command {
 		os.Exit(1)
 	}
 
+	// If PersistentPreRunE returns an error, then cobra displays the error *and*
+	// displays the result of `upgrade-provider --help`. If we don't want help to be
+	// displayed, we can set failedPreRun and return. Run will immediately fail with
+	// this error.
+	var failedPreRun error
 	cmd := &cobra.Command{
 		Use:   "upgrade-provider <provider>",
 		Short: "upgrade-provider automates the process of upgrading a TF-bridged provider",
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):
Yay!

PR #136, opened 2023-09-08
Skip upstream upgrade when target = current

iwahbe:
This case was handled when `ctx.InferVersion`, but not handled correctly when the upstream
version came directly from GitHub.

### DIFF
--- a/upgrade/steps-helpers.go
+++ b/upgrade/steps-helpers.go
@@ -436,19 +436,19 @@ func getRepoExpectedLocation(ctx Context, cwd, repoPath string) (string, error) 
 // sorted by semantic version. The list may be empty.
 //
 // The second argument represents a message to describe the result. It may be empty.
-func GetExpectedTarget(ctx Context, name, upstreamOrg string) (*UpstreamUpgradeTarget, string, error) {
+func GetExpectedTarget(ctx Context, name, upstreamOrg string) (*UpstreamUpgradeTarget, error) {
 	// InferVersion == true: use issue system, with ctx.TargetVersion limiting the version if set
 	if ctx.InferVersion {
 		return getExpectedTargetFromIssues(ctx, name)
 	}
 	if ctx.TargetVersion != nil {
-		return &UpstreamUpgradeTarget{Version: ctx.TargetVersion}, "", nil
+		return &UpstreamUpgradeTarget{Version: ctx.TargetVersion}, nil
 
 	}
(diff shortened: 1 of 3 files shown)


upgrade/steps-helpers.go
```diff
@@ -506,33 +505,26 @@ func getExpectedTargetFromIssues(ctx Context, name string) (*UpstreamUpgradeTarg
...
+		return nil, nil
 	}
 	sort.Slice(versions, func(i, j int) bool {
 		return versions[j].Version.LessThan(versions[i].Version)
 	})
 
-	if ctx.InferVersion && ctx.TargetVersion != nil && !versions[0].Version.Equal(ctx.TargetVersion) {
-		return nil, "", fmt.Errorf("possible upgrades exist, but non match %s", ctx.TargetVersion)
+	if ctx.TargetVersion != nil && !versions[0].Version.Equal(ctx.TargetVersion) {
+		return nil, fmt.Errorf("possible upgrades exist, but non match %s", ctx.TargetVersion)
```
mjeffryes, line 524: `s/non/none/` 

Also I'm unclear why we're only testing against `versions[0]` rather than searching `versions` for a match?

upgrade/steps-helpers.go
```diff
@@ -506,33 +505,26 @@ func getExpectedTargetFromIssues(ctx Context, name string) (*UpstreamUpgradeTarg
...
+			continue
 		}
+		versions = append(versions, UpgradeTargetIssue{
+			Version: v,
+			Number:  title.Number,
+		})
 	}
 	if len(versions) == 0 {
-		var extra string
-		if ctx.InferVersion && versionConstrained {
```
mjeffryes, line 522: is `ctx.InferVersion` used anywhere else? If not, should we get rid of it?

upgrade/steps-helpers.go
iwahbe: We should be doing that. Fixed in https://github.com/pulumi/upgrade-provider/pull/136/commits/2d2e40c4b4cc0e849ee639331e95f22018ccbc56.

upgrade/steps-helpers.go
iwahbe: We still use it: https://github.com/pulumi/upgrade-provider/blob/2d2e40c4b4cc0e849ee639331e95f22018ccbc56/upgrade/steps-helpers.go#L441.

mjeffryes (approved):
I don't have full context here, but I think these changes make sense.

PR #137, opened 2023-09-08
Upgrade plugin sdk after bridge upgrade

iwahbe:
Attempt to upgrade `pulumi/terraform-plugin-sdk` after each bridge upgrade. We do this because the bridge might use new features of `pulumi/terraform-plugin-sdk`, thus failing to compile on old versions.

Stacked on top of https://github.com/pulumi/upgrade-provider/pull/136.


aq17 (approved):
LGTM

PR #141, opened 2023-09-12
Always run the TF Plugin SDK Upgrader

iwahbe:
We recently moved this from running during provider upgrades to running during bridge upgrades. We should actually upgrade in either case, since both the provider and the bridge can rely on a new version of the SDK.

---

In the past, we duplicated work; calculating the version to use in a planning step and then recalculating the SHA in the upgrade step itself. This PR calculates the SHA directly in the planning step, and then uses that in the upgrade step.

---

Example PR is https://github.com/pulumi/pulumi-aws/pull/2804.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -338,43 +338,10 @@ func InformGitHub(
 //
 // This is predicated on updating to the latest version being safe. We will need to
 // revisit this when a new major version of the plugin SDK is released.
-func getLatestTFPluginSDKReplace(ctx context.Context, repo ProviderRepo, targetSHA *string) step.Step {
+func setTFPluginSDKReplace(ctx context.Context, repo ProviderRepo, targetSHA *string) step.Step {
 	// We do discover in a step.Computed so if the fork isn't present, it isn't
 	// displayed to the user.
 	return step.F("Update TF Plugin SDK Fork", func() (string, error) {
-		// If the fork is present, we need to figure out the SHA of the
-		// latest upstream version to use.
-		const hostRepo = "https://github.com/pulumi/terraform-plugin-sdk.git"
-		result, err := exec.CommandContext(ctx, "git",
-			"ls-remote", "--heads", hostRepo).Output()
-		if err != nil {
-			return "", fmt.Errorf("could not get branches: %w", err)
-		}
-		lines := strings.Split(string(result), "\n")
-		versions := make([]*semver.Version, len(lines))
-		shas := make([]string, len(lines))
-		highest := -1
-		for i, line := range lines {
-			split := strings.Split(strings.TrimSpace(line), "\t")
-			if len(split) < 2 {
-				continue
-			}
-			shas[i] = split[0]
-			version, hasVersion := strings.CutPrefix(split[1], "refs/heads/upstream-")
-			if !hasVersion {
-				continue
-			}
-			if v, err := semver.NewVersion(version); err == nil {
-				versions[i] = v
-				if highest == -1 || versions[highest].LessThan(v) {
-					highest = i
-				}
-			}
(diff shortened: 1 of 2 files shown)


upgrade/steps.go
```diff
@@ -339,116 +338,69 @@ func InformGitHub(
...
+		lines := strings.Split(string(result), "\n")
+		versions := make([]*semver.Version, len(lines))
+		shas := make([]string, len(lines))
+		highest := -1
+		for i, line := range lines {
+			split := strings.Split(strings.TrimSpace(line), "\t")
+			if len(split) < 2 {
+				continue
+			}
+			shas[i] = split[0]
```
guineveresaenger, line 362: do we need shas? I'm not seeing where they're being used, but perhaps I'm mistaken?

upgrade/steps.go
```diff
@@ -339,116 +338,69 @@ func InformGitHub(
...
+			version, hasVersion := strings.CutPrefix(split[1], "refs/heads/upstream-")
+			if !hasVersion {
+				continue
+			}
+			if v, err := semver.NewVersion(version); err == nil {
+				versions[i] = v
+				if highest == -1 || versions[highest].LessThan(v) {
+					highest = i
+				}
```
guineveresaenger, line 371: I feel like I'm missing something here. It looks like we're trying to detect the highest version of the input lines.
But I don't see why, or what we're doing with it.

upgrade/steps.go
iwahbe: Unfortunately, we do. `replace` directives don't take "version queries", they take versions. Since the release we are targeting is the head of the latest branch, we need to insert a `replace` statement pointing to the SHA of that branch HEAD.

upgrade/upgrade-provider.go
```diff
@@ -315,6 +321,17 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 		}()
 	}
 
+	steps = append(steps, step.Computed(func() step.Step {
```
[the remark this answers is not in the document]
iwahbe: We upgrade if and only if we set `tfSDKTargetSHA` to a non `""` value. That only triggers when `ctx.UpgradeBridgeVersion` is set.

upgrade/upgrade-provider.go
```diff
@@ -315,6 +321,17 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
...
+		// No upgrade was planned, so exit
+		if tfSDKTargetSHA == "" {
+			return nil
+		}
+		return step.Combined("Update Plugin SDK",
+			getLatestTFPluginSDKReplace(ctx, repo, &tfSDKTargetSHA),
+			step.Cmd(exec.CommandContext(ctx, "go", "mod", "tidy")).In(repo.providerDir()),
```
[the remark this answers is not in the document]
iwahbe: Yes. `getLatestTFPluginSDKReplace` leaves `go.mod` in a valid but unresolved state. We need to run `go mod tidy` to resolve the SHA into a pseudo version so other `go` commands will work as expected.

We can't even parse `go.mod` when unresolved, since we don't include a resolver in our parsers.

upgrade/steps.go
iwahbe: Your not. We can tear a bunch of this out. Great catch!

t0yv0 (approved):

t0yv0:
LGTM If you tried it out somewhere

iwahbe:
I tried this for the pulumi-aws upgrade PR. https://github.com/pulumi/pulumi-okta/pull/383 is an example of `upgrade-provider` doing nothing (not interfering with anything).

PR #152, opened 2023-09-14
Upgrade pulumi sdk and pkg in examples/ folder when upgrading bridge

### LINKED ISSUE #151: Pulumi SDK should be updated everywhere
When updating the bridge, we should update the pulumi version in `examples/` as well.

guineveresaenger:
Fixes #151

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -809,7 +809,7 @@ func ReplaceAssertNoError(ctx Context, repo ProviderRepo) (step.Step, error) {
 // UpgradePulumiVersions reads the current Pulumi SDK version from go.mod and applies it to:
 // sdk/go.mod
 // examples/go.mod - we also infer the `pkg` version here and add it.
-func BridgePulumiVersions(ctx Context, repo ProviderRepo) (step.Step, error) {
+func BridgePulumiVersions(ctx Context, repo ProviderRepo) step.Step {
 	// When we've updated the bridge version, we need to update the corresponding pulumi version in sdk/go.mod.
 	// It needs to match the version used in provider/go.mod, which is *not* necessarily `latest`.
 	var newSdkVersion string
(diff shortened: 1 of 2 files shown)


upgrade/upgrade-provider.go
```diff
@@ -363,10 +363,32 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
...
+				"get",
+				"github.com/pulumi/pulumi/pkg/v3@"+newSdkVersion)).In(repo.examplesDir())
+		})
+
+		steps = append(steps,
+			getNewPulumiVersionStep,
+			bridgePulumiVersionSdkStep,
+			bridgePulumiVersionExamplesStep,
+			examplesPulumiPkgStep,
+		)
```
iwahbe, line 391: This is sufficiently reparative that it's worth putting in a function

guineveresaenger: What do you mean by "reparative"?

iwahbe: Some combination of "repetitive", autocorrect and bad spelling. My bad.

upgrade/steps.go
```diff
@@ -805,3 +805,54 @@ func ReplaceAssertNoError(ctx Context, repo ProviderRepo) (step.Step, error) {
 	}
 	return step.Combined("Replace AssertNoError with AssertNoErrorf", steps...), nil
 }
+
+// UpgradePulumiVersions reads the current Pulumi SDK version from go.mod and applies it to:
+// sdk/go.mod
+// examples/go.mod - we also infer the `pkg` version here and add it.
+func BridgePulumiVersions(ctx Context, repo ProviderRepo) (step.Step, error) {
```
iwahbe, line 812: I don't see where `BridgePulumiVersions` can return an error.
```suggestion
func BridgePulumiVersions(ctx Context, repo ProviderRepo) step.Step {
```

upgrade/steps.go
```diff
@@ -805,3 +805,54 @@ func ReplaceAssertNoError(ctx Context, repo ProviderRepo) (step.Step, error) {
...
+			"go",
+			"get",
+			"github.com/pulumi/pulumi/pkg/v3@"+newSdkVersion)).In(repo.examplesDir())
+	})
+
+	steps = append(steps,
+		getNewPulumiVersionStep,
+		bridgePulumiVersionSdkStep,
+		bridgePulumiVersionExamplesStep,
+		examplesPulumiPkgStep)
```
iwahbe, line 855: I find this much easier to parse, since we make it clear that each upgrade is the same, parameterized over the mod path:
```suggestion
	upgrade := func(mod string) step.Step {
		return step.Computed(func() step.Step {
			return step.Cmd(exec.CommandContext(ctx,
				"go", "get",
				"github.com/pulumi/pulumi/"+mod+"/v3@"+newSdkVersion))
		})
	}

	steps = append(steps,
		getNewPulumiVersionStep,
		upgrade("sdk").In(repo.sdkDir()),
		upgrade("sdk").In(repo.examplesDir()),
		// In examples/ folder, we use pulumi/pkg as well. We should keep this in sync also.
		upgrade("pkg").In(repo.examplesDir()))
```

This is what I meant by moving repetitive code into a function.

upgrade/steps.go
iwahbe: I love moving this concept into a function.

guineveresaenger: It's returning an error here; https://github.com/pulumi/upgrade-provider/pull/152/files#diff-d5c54fe7fdac4cea34363fc36154d215683b9b55e14428dd7cd578204f6c8d92R820

upgrade/steps.go
guineveresaenger: Yes, I tried something like that. And again, I cannot figure out how to make steps.Comupted() do the right thing - the `newSDKVersion` doesn't get evaluated.

iwahbe (approved):
LGTM!

PR #153, opened 2023-09-14
Correctly map git tags to shas for tag discovery

### LINKED ISSUE #150: `gitRefsOf` overwrites tags when the SHA is tagged multiple times
### What happened?

During `upgrade-provider --kind=pf`, I got the following output:

✓ Repo kind: patched
✓ Planning Plugin Framework Update: Up to date at 0.15.2

This is not accurate - latest is at 0.16.0.

### Expected Behavior

Upgrade to v0.16.0, which is the latest pf tag.

### Steps to reproduce

Check out Akamai and try to run a bridge upgrade with kind==pf

### Output of `pulumi about`

n/a

### Additional context

This is happening [here](https://github.com/pulumi/upgrade-provider/blob/main/upgrade/steps-helpers.go#L362) when a git sha has _multiple_ tags. The keys in `gitRepoRefs` are not unique, and so tags can get overwritten.

### Contributing

guineveresaenger:
Fixes #150.

- Upgrade tfpf with a bridge update always
- These changes work on the pf-only upgrade at least

### DIFF
--- a/upgrade/steps-helpers.go
+++ b/upgrade/steps-helpers.go
@@ -350,7 +350,7 @@ func gitRefsOf(ctx context.Context, url, kind string) (gitRepoRefs, error) {
 	if err := cmd.Run(); err != nil {
 		return gitRepoRefs{}, fmt.Errorf("git %s: %w", strings.Join(args, " "), err)
 	}
-	refsToBranches := map[string]string{}
+	branchesToRefs := map[string]string{}
 	for i, line := range strings.Split(out.String(), "\n") {
 		if line == "" {
 			continue
@@ -359,18 +359,19 @@ func gitRefsOf(ctx context.Context, url, kind string) (gitRepoRefs, error) {
 		contract.Assertf(len(parts) == 2,
 			"expected git ls-remote to give '\t' separated values, found line %d: '%s'",
 			i, line)
-		refsToBranches[parts[0]] = parts[1]
+		branchesToRefs[parts[1]] = parts[0]
 	}
-	return gitRepoRefs{refsToBranches, kind}, nil
+	return gitRepoRefs{branchesToRefs, kind}, nil
+
 }
 
 type gitRepoRefs struct {
-	refsToLabel map[string]string
-	kind        string
+	LabelToRef map[string]string
+	kind       string
 }
 
 func (g gitRepoRefs) shaOf(label string) (string, bool) {
-	for ref, l := range g.refsToLabel {
+	for l, ref := range g.LabelToRef {
 		if l == label {
 			return ref, true
 		}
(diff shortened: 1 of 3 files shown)


upgrade/util.go
```diff
@@ -144,6 +144,7 @@ func latestSemverTag(prefix string, refs gitRepoRefs) *semver.Version {
 			return vA.GreaterThan(vB)
 		}
 	})
+	//fmt.Println("sorted", sorted)
```
iwahbe, line 147: Debug?

upgrade/steps-helpers.go
```diff
@@ -359,18 +359,19 @@ func gitRefsOf(ctx context.Context, url, kind string) (gitRepoRefs, error) {
...
 	}
-	return gitRepoRefs{refsToBranches, kind}, nil
+	return gitRepoRefs{branchesToRefs, kind}, nil
+
 }
 
 type gitRepoRefs struct {
-	refsToLabel map[string]string
-	kind        string
+	LabelToRef map[string]string
```
iwahbe, line 369: Why did we change capitalization here?
```suggestion
	labelToRef map[string]string
```

upgrade/upgrade-provider.go
```diff
@@ -331,6 +331,10 @@ func UpgradeProvider(ctx Context, repoOrg, repoName string) error {
 			"go", "get", "github.com/pulumi/pulumi-terraform-bridge/v3@"+targetBridgeVersion)).
 			In(repo.providerDir()))
 
+		steps = append(steps, step.Cmd(exec.CommandContext(ctx,
+			"go", "get", "github.com/hashicorp/terraform-plugin-framework")).
+			In(repo.providerDir()))
```
iwahbe, line 336: What if the provider doesn't use (or compile against) the latest version of the bridge?

iwahbe (commented):
1 small nit on the sha tags, but that looks good.

I'm more nervous about the **ADDITIONALLY**.

upgrade/upgrade-provider.go
guineveresaenger: Have we encountered such a situation, especially when upgrading the bridge explicitly? if so, then perhaps there's a guard flag somewhere we can use?

upgrade/steps-helpers.go
guineveresaenger: No reason; I can flip it back.

guineveresaenger:
Okay, let's make separate PRs for these. Will rebase.

upgrade/upgrade-provider.go
iwahbe: I think we saw that for TLS. I'm generally nervous about adding to many guard flags. It makes the tool very hard to use.

iwahbe (approved):
LGTM!

PR #170, opened 2023-10-02
Pass `context.Context` in steps.

iwahbe:
I'm going to try to start refactoring the step.Step library so we can use it to generate
replay tests. That will require contextual information, so I have made the decision to
pass context.Context now.

Right now that means:
- Changing `step.F` to take a `func(context.Context) (string, error)`
- Changing `step.Cmd` to take `name string, ...args` instead of an `exec.Command`.

This change will have no user facing behavior.

### DIFF
--- a/step/step.go
+++ b/step/step.go
@@ -5,6 +5,7 @@
 package step
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


iwahbe:
Tested in https://github.com/pulumi/pulumi-venafi/pull/180.

step/step.go
```diff
@@ -271,10 +274,22 @@ func (c combined) run(prefix string) bool {
 	return true
 }
 
+func Value(desc, s string) Step {
```
guineveresaenger, line 277: Can we have a comment here what Value does/when to use it? sounds like it's for returning the outcome of a step?

guineveresaenger (approved):

step/step.go
iwahbe: Sure

PR #174, opened 2023-10-11
Hide *Context inside context.Context

iwahbe:
This allows multiple systems to pass context.Context without interference, normalizing function signatures. This is a pure refactor and does not have user facing changes.


t0yv0 (approved):
Improved Go style also, appreciated.

PR #176, opened 2023-10-13
Extensible PR description

t0yv0:
Part of enabling better bridge downstream checks.

Example PR generated with this version:

https://github.com/pulumi/pulumi-aiven/pull/395

```
(cd ~/code/pulumi-aiven && ~/code/upgrade-provider/upgrade-provider \
    pulumi/pulumi-aiven \
    --kind=bridge \
    --target-bridge-version=$(cd ~/code/pulumi-terraform-bridge && git rev-parse HEAD) \
    --pr-reviewers "t0yv0" \
    --pr-description "This is some more PR description linking back to https://github.com/pulumi/pulumi-terraform-bridge/pull/1408 that is also expressible as pulumi/pulumi-terraform-bridge#1408")

```


t0yv0:
I'm wondering why reviewers takes no effect. If I specify "@t0yv0" it fails with exception. If I specify "t0yv0" it leaves no reviewers which is good (compared to polyfilling providers team) but it does not tag me.

t0yv0:
I guess I should remove the double-stating of the description with --via it's not pretty.

t0yv0:
Actually I managed to add an end-to-end test https://github.com/pulumi/pulumi-aiven/pull/398 with ci-mgmt and the wrapper action and everything. SO far so good. It's able to  open PR with extra description, custom PR reviewers and build against a hash version of the bridge.

mjeffryes (approved):

PR #181, opened 2023-10-17
New step library

iwahbe:
The `upgrade-provider` tool has grown from a binary to be run locally to make upgrading providers easier to a vital part of our automation and a publicly available tool for other maintainers of bridged providers.

It was originally designed as a side project not meant for production use. Because it would only be run locally by its author, there was no need to test. This no longer applies. 

Other developers have struggled with aggressively asynchronous native of the `step` library. This is further aggravated by the lack of need for asynchrony. 

`step/v2` aims to solve both of these problems: It provides a synchronous, strongly typed way to annotate functions. Annotated functions are displayed to the user. Unlike `step` (v1), annotated functions are just normal functions. They can call other functions, with or without annotations. They run synchronously and return normal values.

## Example Pipeline

```go
func topLevel(ctx context.Context) {
	a, err := strconv.Atoi(step.ReadFile(ctx, "a.txt")) // a is a string
	step.HaltOnError(ctx, err)
	b, err := strconv.Atoi(step.ReadFile(ctx, "b.txt")) // b is a string
	step.HaltOnError(ctx, err)
	result := step.Call21(ctx, "plus", plus, a, b) // result is a number
	step.WriteFile(ctx, "c.txt", fmt.Sprintf("%d\n", result))
}

// step can call any function as an annotated function, but it must
// take a context.Context as its first argument.
func plus(_ context.Context, a, b int) int { return a + b }

func main() {
	err := step.Pipeline("sum files", topLevel)
	if err != nil {
		fmt.Printf("error: %s\n", err.Error())
	}
}
```

Note: `step.ReadFile`, `step.WriteFile` and `step.HaltOnError` are helper functions. They don't access any internals of the `step` library.

To add replay capabilities to the pipeline, we need to add a recorder. We will do this when an env var is set, just like our other tools:

```go
func main() {
    ctx := context.Background()
    if file := os.Getenv("PULUMI_REPLAY"); file != "" {
	var write io.Closer
	ctx, write = step.WithRecord(ctx, file)
	defer func() { if err := write.Close(); err != nil { panic(err) } }()
    }
    err := step.Pipeline(ctx, "sum files", topLevel)
    if err != nil { fmt.Printf("error: %s\n", err.Error()) }
}
```

Now we can record our pipeline run.
```
$ PULUMI_REPLAY=logs.json example
```

This will write our replay to `logs.json`:
```json
{
  "pipelines": [{
      "name": "sum files",
      "steps": [{
          "name": "a.txt",
          "inputs": ["a.txt"],
          "outputs": ["1", null],
          "impure": true
        },
        {
          "name": "b.txt",
          "inputs": ["b.txt"],
          "outputs": ["2", null],
          "impure": true
        },
        {
          "name": "plus",
          "inputs": [1, 2],
          "outputs": [3, null]
        },
        {
          "name": "c.txt",
          "inputs": ["c.txt", "3\n"],
          "outputs": [null],
          "impure": true
        }]
    }]
}
```

Unit tests can include pre-recorded copies of `logs.json`, asserting that the pipeline follows the recorded steps. Note that certain steps are marked `"impure": true`. This indicates that the step is not safe or reasonable to run in an isolated test environment. When we run the replay test, the provider will assert that we called `step.ReadFile` and `step.WriteFile` with the correct inputs, and will return the recorded outputs. 

Finally, observe that `step.ReadFile` returns a `string`, not a `(string, error)`. `step/v2` provides error handling, halting a pipeline when an error is encountered and displaying it to the user. This makes our code a lot easier to read when many pipelines consist of a series of commands.

### Pipelines

`step/v2` forms pipelines, each of which consist of a series of steps. Both steps and pipelines are named. Pipelines can be created with `step.Pipeline` or `step.PipelineCtx`. Within pipelines, steps can be called.

### Steps

The library generates generic `step.Call` functions so you can call your own annotated functions. Each generated function looks like `step.CallIO(E?)` where `I` is the number of input parameters, `O` is the number of output parameters, and `E` indicates that an error is at the end. Every annotated function takes a `context.Context`, which is not counted in the number of inputs. For example:

```go
func sum(_ context.Context, a, b int) int { return a + b }
result := step.Call21(ctx, "Sum", sum, 1, 2)
result == 3 // true
```

Since `sum` has two inputs (beyond `context.Context`) and one output, we call it with `step.Call21`.

### Annotations

There are a couple of built in annotations that annotated functions can call.

`SetLabel` sets text that displays next to the annotated function. Borrowing from out previous sum example, usage could look like:
```go
func sum(ctx context.Context, a , b int) int {
    step.SetLabel(ctx, fmt.Sprintf("Adding %d + %d", a, b))
    result := a + b
    step.SetLabel(ctx, fmt.Sprintf("%d + %d = %d", a, b, result))
    return result
}
```

Each call to `step.SetLabel` will override the previous call in the user's display.

The other important annotation to understand is `MarkImpure`. This indicates to replay tests that the function will have side effects or depend on external state, and so should not be run in a test context. Usage looks like this:

https://github.com/pulumi/upgrade-provider/blob/2d789477512e2dc4b5c2c3d1161340d43d01c31b/step/v2/util.go#L28-L34

This is how the replay tests in our original example work without `a.txt` and `b.txt` being present while testing.

## Migration from `step` (v1)

As we touch various parts of the code base and want to add tests, we will migrate to the new framework. I expect that I (@iwahbe) will be doing most of this as part of driving the issue count on this repo to 0. https://github.com/pulumi/upgrade-provider/pull/182 shows that it doesn't take much effort to migrate.


step/v2/step.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil
+	}
+	return v.(*pipeline)
+}
+
+func Pipeline(name string, steps func(context.Context)) error {
+	return PipelineCtx(context.Background(), name, steps)
+}
+
+func PipelineCtx(ctx context.Context, name string, steps func(context.Context)) (err error) {
```
mjeffryes, line 50: nit. add comments to describe public methods

step/v2/step.go
```diff
@@ -0,0 +1,264 @@
...
+			p.spinner.Prefix, p.failed.Error())
+	}
+	p.spinner.Stop()
+	return p.failed
+}
+
+func mustGetPipeline(ctx context.Context, name string) *pipeline {
+	p := getPipeline(ctx)
+	if p == nil {
+		panic(`Must call "` + name + `" on a context from a step function`)
```
mjeffryes, line 98: Isn't it more that it needs to be called in the context of a pipeline? Also, I wonder if it's really necessary to hard fail here. You should be able to run the steps without a pipeline, you just lose the functionality the pipeline adds.

step/v2/examples/simple/main.go
```diff
@@ -0,0 +1,63 @@
...
+		}()
+	}
+
+	if err := step.PipelineCtx(ctx, "simple", pipeline); err != nil {
+		os.Exit(1)
+	}
+}
+
+func pipeline(ctx context.Context) {
+	bytes := step.ReadFile(ctx, "input.txt")
```
mjeffryes, line 32: Suppose you wanted to handle an error (eg. try to open this file, if it fails open this other file) how would you do it?

step/v2/step.go
```diff
@@ -0,0 +1,264 @@
...
+
+func getPipeline(ctx context.Context) *pipeline {
+	v := ctx.Value(pipelineKey{})
+	if v == nil {
+		return nil
+	}
+	return v.(*pipeline)
+}
+
+func Pipeline(name string, steps func(context.Context)) error {
```
mjeffryes, line 46: This may not be worth the effort to refactor, but I notice that pipelines are handling multiple concerns (eg. fancy display updates, recording a trace for replay). I think ideally these would be composble. So you could run a pipeline with tracing only or a pipeline with fancy display only or one with both, or you could swap out a different display feature that uses bubbletea, etc. 

Again, probably not worth spending a ton of time on it at the moment, but if it's not a difficult refactor, it would make the framework more flexible.

mjeffryes (commented):
This is fun. As a total nerd I love reviewing this kind of tool. 

I do wonder if there's not something off the shelf we should be using for this; it doesn't seem like we would be the first people to need to define a pipeline of steps and make fancy spinners and record a trace.

Most of my comments are not blocking, but I do think you might want to plan for error handling before landing.

t0yv0:
Free monads come to Golang. This is fun indeed.

t0yv0:
I'm engaging lightly here I think fundamentally thank you very much, some form of testing is sorely needed and doing casual contributions here and not being able to symbolically test them has caused enough friction for me that I'm really grateful for introducing that whatever form it may take.

On the design side this probably looks like slightly non-idiomatic Go but I wouldn't block on it.

It sounds like you have two problems, you need to isolate side-effects and you need to decorate functions. Every language has approaches to these things, hopefully one canonical one but Haskell people can't agree on free monads vs effects vs TML. So what does it look like in Go? I'm not sure TBH:

Dependency injection via Context sounds legitimate; you could also manually propagate interface references for any effects your program is doing. At scale people use https://github.com/uber-go/dig but this codebase does not have  scale.

For function decorators Python has the lead here of making it nice and easy so people stop arguing how to do it. The direct way that seems to be happening here is probably fine, I'm guessing: 

```golang
func MyFunc(_ Context, x int, y string) (bool, error) {
  ..
}
// =>

var MyFunc = decorate(func (_ Context, x int, y string) (bool, error) {
  ..
})

```

Sad that it makes MyFunc mutable but gentlefolk do NOT mutate.  

Something like this could be made to work but is more cumbersome. This is the approach I've seen taken for tracing decorators (performance).

```golang
func MyFunc(ctx Context, x int, y string) (z bool, e error) {
  d := decorate(ctx, x, y) 
  defer d.done(z, e)
  ..
}
```

step/v2/examples/simple/main.go
iwahbe: You can't do that with `step.ReadFile`, but `step.ReadFile` isn't special. It's implementation looks like this:

https://github.com/pulumi/upgrade-provider/blob/2d789477512e2dc4b5c2c3d1161340d43d01c31b/step/v2/util.go#L18-L26

Because go `error`s are just values, `ReadFileOrError` would look like this:

```go
// Read a file, halting the pipeline on error.
func ReadFileOrError(ctx context.Context, path string) (string, error) {
	return Call12(ctx, path, func(ctx context.Context, path string) (string, error) {
		MarkImpure(ctx)
		bytes, err := os.ReadFile(path)
		SetLabel(ctx, fmt.Sprintf("%d bytes read", len(bytes)))
		return string(bytes), err
	}, path)
}
```

step/v2/step_test.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+func TestCallTree(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		stack    []string
+		expected string
+	}{
+		{
+			stack: []string{"foo", "bar", "fizz", "", "", "bar2"},
```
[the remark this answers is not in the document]
iwahbe: This is a relatively simple way to define a tree structure *and* a walkable cursor into that tree structure. It seemed like a lot less code then defining a full tree structure with backlinks, and then traversing that.

step/v2/examples/simple/main.go
mjeffryes: Ah, ok, thanks for clarifying; possibly worth exposing both versions of steps in our "stdlib", but seeing this resolves my concern about a structural limitation

step/v2/step_test.go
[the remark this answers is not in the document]
mjeffryes: Ok, I would have probably leaned the other way, but seems like it could evolve later if needed.

mjeffryes (approved):
There's lots of ways to bikeshed here, I'm going to approve since I think this version is reasonable top ship. And will leave to your judgement how much of the feedback you want to incorporate.

step/v2/examples/simple/main.go
mjeffryes: Ah, ok, thanks for clarifying; possibly worth exposing both versions of steps in our "stdlib", but seeing this resolves my concern about a structural limitation

step/v2/examples/simple/main.go
```diff
@@ -0,0 +1,63 @@
...
+	process := step.Call11(ctx, "hide-secret", hide, bytes)
```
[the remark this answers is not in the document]
iwahbe: I like it. Fixed in https://github.com/pulumi/upgrade-provider/pull/181/commits/0891e441a96ce90c0d0b51121ebac7399a90cd7b.

PR #182, opened 2023-10-17
Rewrite inform GitHub

iwahbe:
This PR rewrites the final section of the upgrade provider tool to use the `step/v2` library. Rewriting in `step/v2` has multiple benefits:

- Unlike `step` or the tool itself, `step/v2` was built with testing in mind: I have included a replay test for `InformGitHub` that asserts that if `InformGitHub` is called with the described inputs, it will call the correct set of shell commands with the correct set of arguments.

- `step/v2` is synchronous and strongly typed. Calling a step to read a file looks like this:
  https://github.com/pulumi/upgrade-provider/blob/da557c51f6c45f5468894c13c6aa35e684471aad/upgrade/steps.go#L699
  `content` is then just a string. It doesn't need special handling and you can read it immediately.

For a stronger description of the difference between `step` and `step/v2`, please see https://github.com/pulumi/upgrade-provider/pull/181.

### DIFF
--- a/step/v2/util.go
+++ b/step/v2/util.go
@@ -54,3 +54,19 @@ func HaltOnError(ctx context.Context, err error) {
 		return err
 	})(ctx)
 }
+
+// Run `f` in `dir`.
+//
+// If the dir does not exist, then the current pipeline will fail.
+func WithCwd(ctx context.Context, dir string, f func(context.Context)) {
+	c := &Cwd{To: dir}
+	err := c.Enter(ctx, StepInfo{})
+	HaltOnError(ctx, err)
+
+	defer func() {
+		err := c.Exit(ctx, nil)
+		HaltOnError(ctx, err)
+	}()
+
+	f(WithEnv(ctx, c))
+}
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -20,23 +20,19 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/pulumi/upgrade-provider/step"
+	stepv2 "github.com/pulumi/upgrade-provider/step/v2"
 )
 
 // A "git commit" step that is resilient to no changes in the directory.
-// ß
+//
 // This is required to accommodate failure and retry in the `git` push steps.
-func GitCommit(ctx context.Context, msg string) step.Step {
-	return step.Computed(func() step.Step {
-		check, err := exec.CommandContext(ctx, "git", "status", "--porcelain=1").CombinedOutput()
-		description := fmt.Sprintf(`git commit -m "%s"`, msg)
-		if err != nil {
-			return step.Error("git commit", err)
-		}
-		if len(check) > 0 {
-			return step.Cmd("git", "commit", "-m", msg)
-		}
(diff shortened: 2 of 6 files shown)


step/v2/util.go
```diff
@@ -35,16 +35,14 @@ func WriteFile(ctx context.Context, path, content string) {
...
 		MarkImpure(ctx)
-		out, err := cmd.Output()
+		out, err := exec.CommandContext(ctx, name, args...).Output()
 		if exit, ok := err.(*exec.ExitError); ok {
 			err = fmt.Errorf("%s:\n%s", err.Error(), string(exit.Stderr))
 		}
 		return string(out), err
-	}, key)
+	}, name, args)
 }
```
mjeffryes, line 47: I take it you'll fold this into #181 before landing?

upgrade/steps.go
```diff
@@ -307,9 +303,11 @@ func UpgradeProviderVersion(
 func InformGitHub(
 	ctx context.Context, target *UpstreamUpgradeTarget, repo ProviderRepo,
 	goMod *GoMod, targetBridgeVersion, targetPfVersion Ref, tfSDKUpgrade string,
-) step.Step {
-	pushBranch := step.Cmd("git", "push", "--set-upstream",
-		"origin", repo.workingBranch).In(&repo.root)
+	osArgs []string,
+) error {
+	ctx = stepv2.WithEnv(ctx, &stepv2.Cwd{To: repo.root})
```
mjeffryes, line 308: I don't love this syntax for setting the working directory. At minimum, it would be nice to have a utility function, but honestly, I'd rather somehow wrap the function(s) that's executing with a CWD, to make it really obvious where it applies.

upgrade/steps.go
```diff
@@ -324,38 +322,32 @@ func InformGitHub(
...
-			}
-			return step.Combined("Assign Issues", issues...)
-		}),
-	)
+	// If we are only upgrading the bridge, we wont have a list of issues.
+	if !GetContext(ctx).UpgradeProviderVersion {
+		return nil
+	}
+
+	stepv2.Call00(ctx, "Assign Issues", func(ctx context.Context) {
```
mjeffryes, line 341: It kind of seems like `Call00` should have a special name; iiuc, this is just providing a grouping with a label for the steps inside right?

mjeffryes (commented):
Nice! I like the new tests and it does seem like the v2 steps have fewer warts in the UX

step/v2/util.go
iwahbe: That's the plan.

upgrade/steps.go
iwahbe: What do you think about this?
```go
pwd = stepv2.Cmd(ctx, "pwd") // pwd ends with /base
stepv2.WithCwd(ctx, "./dir, func(ctx context.Context) {
     pwd = stepv2.Cmd(ctx, "pwd") // pwd ends with base/dir
})
pwd = stepv2.Cmd(ctx, "pwd") // pwd ends with /base again
```

PR #183, opened 2023-10-17
Disable PF upgrade when /pf is not found, not UpgradeProviderVersion

### LINKED ISSUE #179: "provider" mode does not work
### What happened?

After [context refactor pr](https://github.com/pulumi/upgrade-provider/pull/174) I cannot upgrade in provider mode anymore.
Checked that the problem still exists in the [latest commit](https://github.com/pulumi/upgrade-provider/commit/4791ea6890c04b8540efe19161ec13e09085b495)
Also checked [one commit before the PR](https://github.com/pulumi/upgrade-provider/pull/171) - it works as expected.

Could you please check what the magic happed there? Why do I see pending update and then it's ignored. I tried to dig in the code... looks like it's some references bug.

image

cc: @iwahbe 

### Example

https://github.com/pulumiverse/pulumi-zitadel/actions/runs/6546386505/job/17776740556#step:10:27

### Output of `pulumi about`

It's inside of CI. I'm using this action to setup the latest pulumi 
https://github.com/pulumiverse/pulumi-zitadel/blob/6bb78c4b2a3b08a5fd2437ab67f18a7da684491a/.github/workflows/upgrade-provider.yml#L28

### Additional context

_No response_

### Contributing

iwahbe:
This regression was introduced in https://github.com/pulumi/upgrade-provider/pull/174. Before #174, `Context` was passed by value, and calling `setFalse` didn't do anything. I believe that code was always broken, but since it mutated a copy it was harmless.

This is a priority fix, since our automation is currently broken for providers that don't use /pf.

Fixes #179

### DIFF
--- a/upgrade/kind.go
+++ b/upgrade/kind.go
@@ -104,9 +104,11 @@ func GetRepoKind(ctx context.Context, repo ProviderRepo) (*GoMod, error) {
 	pf, ok, err := originalGoVersionOf(ctx, repo, filepath.Join("provider", "go.mod"), "github.com/pulumi/pulumi-terraform-bridge/pf")
 	if err != nil {
 		return nil, err
-	}
-	if ctx := GetContext(ctx); !ok && ctx.UpgradeProviderVersion {
-		setFalse(&ctx.UpgradeProviderVersion)
+	} else if !ok {
+		// If we successfully opened provider/go.mod but didn't find any reference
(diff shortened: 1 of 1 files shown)


iwahbe:
I have checked that we didn't have any previous copy -> live changes in this file.

mjeffryes (approved):

PR #189, opened 2023-10-25
Keep the sdk version inline with the bridge

### LINKED ISSUE #184: Do not upgrade pulumi/pkg and pulumi/sdk ahead of the bridge
## Hello!

## Issue details

This is contrary to prior requests originating from me, but I believe at the moment our deployment pipeline would benefit from never upgrading pulumi/pkg references (which installs codegen fixes) ahead of what was tested in pulumi-terraform-bridge. This makes sure additional bridge release checks such as [this](https://github.com/pulumi/pulumi-terraform-bridge/actions/workflows/update-providers-test.yml) qualify the codegen changes on actual provider test-suites.

iwahbe:
This removes the `--kind=sdk` option from the tool.

We already have functionality to bring the pkg+sdk version in all places up to what the bridge requires.

Fixes #184

### DIFF
--- a/main.go
+++ b/main.go
@@ -131,18 +131,6 @@ func cmd() *cobra.Command {
 					if experimental {
 						context.UpgradeCodeMigration = true
 					}
-
-					if len(upgradeKind) > 1 {
-						for _, v := range upgradeKind {
-							if v == "sdk" {
-								context.UpgradeSdkVersion = true
-							}
-						}
-						if !context.UpgradeSdkVersion {
-							warnedAll = true
-							warn("`--kind=all` implies all of bridge,provider,code options")
-						}
-					}
 				case "bridge":
 					set(&context.UpgradeBridgeVersion)
 					set(&context.UpgradePfVersion)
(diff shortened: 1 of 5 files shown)


t0yv0:
I'm a little confused here as it removes the ability to update SDK completely. Is that too much? Does it update the SDK as part of the "bridge" or "all" update? perhaps just dropping that from "all" suffices?

iwahbe:
I thought about dropping `sdk` from `all`, but leaving it as an explicit option. I couldn't think of a situation where we want to upgrade the SDK (which only goes to latest) to differ from the bridge version, or where we would recommend downstream users to do so.

It's not a safe upgrade, since sometimes `pu/pkg` makes breaking changes and the bridge needs to adjust. `upgrade-provider --kind=sdk` may fail, even when every party adheres to their contract. `upgrade-provider --kind=bridge` should not fail if every sub-system works as expected.

t0yv0:
Hm, that's reasonable but how about upgrading native providers that have no bridge?  I guess we can deal with that when we need to. I was also thinking to consider to make it triple-certain, basically "ensure", the invariant that {pkg,sdk} dependency is strictly never ahead (or is equal) to the bridge's {pkg,sdk} dependency. But this automation can get in the way of mowing down custom go.mod setups.

t0yv0 (approved):

iwahbe:
This tool really doesn't work on native providers (calling `make tfgen`, etc). They don't have a standardized layout, so can't have a standardized upgrade process.

PR #190, opened 2023-10-25
WIP Begin to convert Discovery to stepv2

iwahbe:
This is a pure refractor for testability. There should be no user facing changes.


step/v2/util.go
```diff
@@ -74,3 +74,23 @@ func WithCwd(ctx context.Context, dir string, f func(context.Context)) {
...
+		if err == nil {
+			SetLabel(ctx, c)
+		}
+		return c, err
+	})(ctx)
+}
+
+// MkDirAll wraps os.MkdirAll with error handling and impure validation.
+func MkDirAll(ctx context.Context, path string, perm os.FileMode) {
+	Func20E("MkDirAll", func(_ context.Context, path string, perm os.FileMode) error {
```
rquitales, line 92: Aside: Is there documentation I can read up on for all these `Func*` functions? Makes it much harder to review with little context.

rquitales (commented):
Changes looks like they roughly follow the same steps as before. I'll admit it was a bit tough to review initially with all the `stepv2.FuncXX` functions making readability difficult. I assume these are pathways are already tested?

iwahbe:
There is testing for `ensureUpstreamRepo`. There isn't testing for `ensureUpstreamFork`, but that is only used by vault. There is no less testing now then there was before.

rquitales (approved):

PR #192, opened 2023-10-27
`--target-pulumi-version`: upgrade to a COMMITish pulumi/{pkg,sdk} version

iwahbe:


upgrade/util.go
```diff
@@ -43,6 +43,13 @@ type Context struct {
 	//
 	UpstreamProviderName string
 
+	// The desired version of pulumi/{pkg,sdk} to link to.
+	//
+	// If TargetPulumiVersion is nil, then pulumi/{pkg,sdk} should follow the bridge.
+	//
+	// Otherwise, we will `require` the described version.
+	TargetPulumiVersion string
```
t0yv0, line 51: Could be a Ref.

upgrade/upgrade_provider.go
```diff
@@ -366,29 +366,36 @@ func UpgradeProvider(ctx context.Context, repoOrg, repoName string) (err error)
...
+		steps = append(steps, step.Combined("Upgrade Pf Version",
+			step.Cmd("go", "get",
+				"github.com/pulumi/pulumi-terraform-bridge/pf@"+targetPfVersion.String()),
+			step.Cmd("go", "mod", "tidy"),
+		).In(repo.providerDir()))
+	}
+
+	if v := GetContext(ctx).TargetPulumiVersion; v != "" {
+		steps = append(steps, step.Combined("Upgrade Pulumi Version",
+			step.Cmd("go", "get", "github.com/pulumi/pulumi/pkg/v3@"+v),
```
t0yv0, line 387: AS we just discussed and explored, if V is a pre-release such as a hash, we might need to resort to go mod replace, as we're likely not intending to check that in anyway, and we want to make sure that GO does not automatically pick different versions for indirect deps via go mod tidy.

iwahbe:
Example PR is https://github.com/pulumi/pulumi-aiven/pull/415.

upgrade/upgrade_provider.go
iwahbe: Fixed while you were reviewing. See https://github.com/pulumi/upgrade-provider/pull/192/commits/ac333b22f10f97610864bda7535c46ac722c0ea8.

upgrade/util.go
iwahbe: It can't be `<latest>` and we never do anything  but stringify it.

t0yv0: Sure, no strong opinions. It would let you distinguish if it's a hash or not easily.

iwahbe: Changed to use `Ref`. I don't think we need to do anything special, but I agree it's the correct long term value. May as well take it now.

iwahbe:
Closed in favor of https://github.com/pulumi/upgrade-provider/pull/194.

PR #196, opened 2023-10-31
Prepare upstream upgrades for release

iwahbe:
When we upgrade providers, we always increment the version semver version v${major}.${minor}.${patch} by one. We don't always have the same version as the upstream provider, but we almost always mirror the increment they went with. That is:

If they bumped ${major}, then we will also bump ${major}. If they bumped ${minor}, then we will also bump ${minor}. If they bumped ${patch}, then we will also bump ${patch}.

This is communicated to our automation by adding a label to the PR, which is read during the release. The `upgrade-provider` tool has enough information to apply these labels, so it should do so.

This gets us to 1-click provider upgrades.

### DIFF
--- a/upgrade/steps_test.go
+++ b/upgrade/steps_test.go
@@ -203,3 +203,37 @@ func TestEnsureUpstreamRepo(t *testing.T) {
 	})
 	require.NoError(t, err)
 }
+
+func TestReleaseLabel(t *testing.T) {
+	tests := []struct {
+		from, to string
+		expect   string
+	}{
+		{"v1.2.3", "v1.2.3", ""},
+		{"v1.2.3", "v1.2.4", "needs-release/patch"},
+		{"v1.1.3", "v1.2.4", "needs-release/minor"},
+		{"v2.2.3", "v1.2.4", "needs-release/major"},
+		{"", "v1.2.4", ""},
+		{"v2.2.3", "", ""},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run("", func(t *testing.T) {
+			parse := func(s string) *semver.Version {
+				if s == "" {
+					return nil
+				}
+				return semver.MustParse(s)
+			}
+
+			from, to := parse(tt.from), parse(tt.to)
+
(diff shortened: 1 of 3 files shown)


rquitales (approved):

upgrade/steps_test.go
```diff
@@ -203,3 +203,37 @@ func TestEnsureUpstreamRepo(t *testing.T) {
...
+
+func TestReleaseLabel(t *testing.T) {
+	tests := []struct {
+		from, to string
+		expect   string
+	}{
+		{"v1.2.3", "v1.2.3", ""},
+		{"v1.2.3", "v1.2.4", "needs-release/patch"},
+		{"v1.1.3", "v1.2.4", "needs-release/minor"},
+		{"v2.2.3", "v1.2.4", "needs-release/major"},
```
mjeffryes, line 215: this is kind of an interesting case, because you're actually indicating a downgrade. I wonder if we should actually return nothing in a downgrade just to be safe.

upgrade/testdata/replay/wavefront_inform_github.json
```diff
@@ -79,7 +90,9 @@
               "--title",
               "Upgrade terraform-provider-wavefront to v5.0.5",
               "--body",
-              "This PR was generated via `$ upgrade-provider pulumi/pulumi-wavefront`.\n\n---\n\n- Upgrading terraform-provider-wavefront from 5.0.3  to 5.0.5.\n\tFixes #232\n"
+              "This PR was generated via `$ upgrade-provider pulumi/pulumi-wavefront`.\n\n---\n\n- Upgrading terraform-provider-wavefront from 5.0.3  to 5.0.5.\n\tFixes #232\n",
+              "--label",
+              "needs-release/patch"
```
mjeffryes, line 95: I'm loving the new replay tests.

mjeffryes (approved):

upgrade/steps_test.go
iwahbe: That's a good corner case to test. I like that behavior change.

PR #197, opened 2023-11-14
Adds a nix flake

t0yv0:
This enables a new installation method such as:

```
 devbox global add github:pulumi/upgrade-provider/t0yv0/flakes   
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,28 @@
+{
+  description = "A flake defining upgrade-provider build-from-source package";
+
+  inputs = {
+    nixpkgs.url = github:NixOS/nixpkgs/nixos-23.05;
+  };
+
+  outputs = { self, nixpkgs }: let
+
+    package = { system }: let
+      pkgs = import nixpkgs { system = system; };
+    in pkgs.buildGo120Module rec {
+      name = "upgrade-provider";
+      version = ''${self.rev or "dirty"}'';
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,28 @@
...
+
+    package = { system }: let
+      pkgs = import nixpkgs { system = system; };
+    in pkgs.buildGo120Module rec {
+      name = "upgrade-provider";
+      version = ''${self.rev or "dirty"}'';
+      src = ./.;
+      # subPackages = [ "cmd/pulumictl" ];
+      doCheck = false;
+      vendorSha256 = "sha256-YBteVgcWvIE10ojd9W4grMK8kJ0zXXQiBuEHto3sABI=";
```
mjeffryes, line 18: Will we have to update this any time the source changes?

mjeffryes (approved):
I see no harm in landing this here, but I'm a little concerned it might rot pretty easily.

flake.nix
t0yv0: Yeah, unfortunately this needs to get updated any time go.sum changes. I've not found yet a nix-native way to make his part of the flake.lock to automatically update.

t0yv0: It's not affected by source changes though. When this needs editing, you run `nix build` and it suggests the new sha256 to insert here.

PR #206, opened 2023-11-28
Error on missing mappings

iwahbe:
We should proactively remove missing mappings. In our automated system, that means erroring when they are found.

### DIFF
--- a/upgrade/upgrade_provider.go
+++ b/upgrade/upgrade_provider.go
@@ -25,6 +25,13 @@ var CodeMigrations = map[string]CodeMigration{
 	"assertnoerror": ReplaceAssertNoError,
 }
 
+func setEnv(ctx context.Context, k, v string) {
+	stepv2.Func20E(k+"="+v, func(ctx context.Context, k, v string) error {
+		stepv2.MarkImpure(ctx)
+		return os.Setenv(k, v)
+	})(ctx, k, v)
+}
+
 func UpgradeProvider(ctx context.Context, repoOrg, repoName string) (err error) {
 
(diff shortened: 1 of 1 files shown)


upgrade/upgrade_provider.go
```diff
@@ -44,18 +51,19 @@ func UpgradeProvider(ctx context.Context, repoOrg, repoName string) (err error)
...
 				return "false"
 			}
 			return "true"
-		}()),
-		step.Env("PULUMI_CONVERT_EXAMPLES_CACHE_DIR", ""),
-	))
-	if !ok {
-		return ErrHandled
+		}())
+		env("PULUMI_MISSING_MAPPING_ERROR", "true")
```
guineveresaenger, line 63: I misled you when we spoke earlier. I believe we should instead set `PULUMI_EXTRA_MAPPING_ERROR` to true. Turns out that was the setting that finally tripped on my machine for the AWS upgrade.
https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfgen/generate.go#L1310 is the relevant behavior; the error message checks with what I saw.

If I read [the bridge generate code](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfgen/generate.go) and its commit history correctly, it seems like the bridge errors on a missing mapping by default nowadays, so we probably don't even need to set `PULUMI_MISSING_MAPPING_ERROR` at all.

guineveresaenger (changes requested):
Confusing env vars are named confusingly. I believe we need to set the other one.

guineveresaenger (approved):

PR #207, opened 2023-11-29
Add an extra flag to only check for an upstream version

guineveresaenger:
Add a flag option `check-upstream-provider` to check whether we are on the latest upstream version, and open a "Upgrade terraform-provider-foo" issue on the provider repo in question if none exists.

This functionality is meant to be used in a CI context and will check for upstream provider releases on a cron schedule. 

The created issue will then trigger our existing update workflow. A test PR is [here](https://github.com/pulumi/pulumi-akamai/issues/341) and we can see it triggered the upgrade action accordingly.

Next step will be adding this cron workflow to a few canary providers.

Part of https://github.com/pulumi/ci-mgmt/issues/724 - this step will allow us to remove dependency on https://github.com/pulumi/github-issue-automation. If this performs as expected, followup work will trigger an upgrade workflow directly once CI detects a new version.

### DIFF
--- a/main.go
+++ b/main.go
@@ -200,10 +200,7 @@ If no version is passed, the pulumi/{pkg,sdk} version will track the bridge`)
 - "sdk": Upgrade the Pulumi sdk only.
 - "pf": Upgrade the Plugin Framework only.
 - "code":     Perform some number of code migrations.
-- "check-upstream-version: Determine if we need to upgrade the upstream provider."`)
-	// TODO: hide this
-	//err := cmd.PersistentFlags().MarkHidden("pulumi-infer-version")
-	//contract.AssertNoErrorf(err, "could not mark `pulumi-infer-version` flag as hidden")
+- "check-upstream-version": Determine if we need to upgrade the upstream provider. For use in CI only."`)
 
 	cmd.PersistentFlags().BoolVar(&experimental, "experimental", false,
 		`Enable experimental features, such as auto token mapping and auto aliasing`)
--- a/main.go
+++ b/main.go
@@ -132,8 +132,14 @@ func cmd() *cobra.Command {
 				case "pf":
 					set(&context.UpgradePfVersion)
 				case "check-upstream-version":
+					if targetVersion != "" {
+						return fmt.Errorf(
+							"--kind check-upstream-version is incompatible with --target-version, cannot set both",
+						)
+					}
 					set(&context.UpgradeProviderVersion)
-					set(&context.CheckUpstreamLatest)
+					set(&context.OnlyCheckUpstream)
+
 				default:
 					return fmt.Errorf(
 						"--kind=%s invalid. Must be one of `all`, `bridge`, `provider`, `code`, or `sdk`.",
--- a/main.go
+++ b/main.go
@@ -131,6 +131,9 @@ func cmd() *cobra.Command {
 					set(&context.UpgradeCodeMigration)
 				case "pf":
 					set(&context.UpgradePfVersion)
+				case "check-upstream-version":
+					set(&context.UpgradeProviderVersion)
+					set(&context.CheckUpstreamLatest)
 				default:
 					return fmt.Errorf(
 						"--kind=%s invalid. Must be one of `all`, `bridge`, `provider`, `code`, or `sdk`.",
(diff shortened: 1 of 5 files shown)


t0yv0:
Curious, from an interface perspective, why is this an option and not a new command? Does it interact in some way with the existing functionality of upgrades?

Also nit.. `check-upstream-version` or `check-upstream-provider` which is it? Thanks.

main.go
```diff
@@ -196,7 +199,8 @@ If no version is passed, the pulumi/{pkg,sdk} version will track the bridge`)
 - "provider": Upgrade the upstream provider only.
 - "sdk": Upgrade the Pulumi sdk only.
 - "pf": Upgrade the Plugin Framework only.
-- "code":     Perform some number of code migrations.`)
+- "code":     Perform some number of code migrations.
+- "check-upstream-version": Determine if we need to upgrade the upstream provider. For use in CI only."`)
```
t0yv0, line 203: "For use in CI only" this seems an odd restriction to make? Why is that?

guineveresaenger: The issue check will most likely fail because the pr author we check for is `@me` - so, for CI, pulumi-bot. If you run locally, then the behavior is differnet.

I'm also happy to remove any documentation here altogether and keep this info in our playbooks instead.

guineveresaenger:
@t0yv0 I can be convinced here - the main reason, however, is that ultimately, I'd like to [remove using issues titled a certain way as the trigger for an upgrade altogether](https://github.com/pulumi/ci-mgmt/issues/724#issuecomment-1828568505) and with the goal of using this to ultimately run an actual upgrade, it seemed thematically more appropriate to implement it as an `upgrade --kind`.

upgrade/steps_helpers.go
```diff
@@ -573,3 +576,49 @@ var getExpectedTargetFromIssues = stepv2.Func11E("From Issues", func(ctx context
...
+		Title  string `json:"title"`
+		Number int    `json:"number"`
+	}
+	err = json.Unmarshal(ghSearchBytes.Bytes(), &titles)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal `gh search issues` output: %w", err)
+	}
+
+	// create new issue if none exist
+	if len(titles) == 0 {
```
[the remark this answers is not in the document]
iwahbe: Do you mean this?
```go
	getIssues := exec.CommandContext(ctx, "gh", "search", "issues",
		title,
```

If so, [the experiment I commented above](https://github.com/pulumi/upgrade-provider/pull/207/files/4601f9897dc7ee6ba504651d55a959f6cd38662d#r1412693594) shows that `title` is a fuzzy or contains based search, not an exact one.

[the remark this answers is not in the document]
guineveresaenger: I'd be pretty surprised if "Upgrade terraform-provider-foo to $thisexact version" is insufficient here but if you're genuinely worried we can look for an exact match on `title`.
```
🦉 gh search issues "Upgrade terraform-provider-datadog to v3.32.0" --repo=pulumi/pulumi-datadog --state=open --json=title,number --author=pulumi-bot
[
  {
    "number": 367,
    "title": "Upgrade terraform-provider-datadog to v3.32.0"
  }
]

🦉 gh search issues "Upgrade terraform-provider-datadog to v3.31.0" --repo=pulumi/pulumi-datadog --state=open --json=title,number --author=pulumi-bot
[
  {
    "number": 352,
    "title": "Upgrade terraform-provider-datadog to v3.31.0"
  }
]
```

main.go
```diff
@@ -131,6 +131,15 @@ func cmd() *cobra.Command {
...
 				case "pf":
 					set(&context.UpgradePfVersion)
+				case "check-upstream-version":
+					if targetVersion != "" {
+						return fmt.Errorf(
+							"--kind check-upstream-version is incompatible with --target-version, cannot set both",
+						)
+					}
+					set(&context.UpgradeProviderVersion)
+					set(&context.OnlyCheckUpstream)
```
mjeffryes, line 141: it feels like this actually wants to be two (or even three) context flags: 
- `CheckUpstream` which triggers the logic to look at the upstream for versions
- `CreateIssue` which tries to create the issue if missing (maybe could be combined with `OnlyCheckUpstream` if we'd skip creating the issue when upgrading directly) 
- `OnlyCheckUpstream` which causes the command to return early after issue creation

mjeffryes (approved):
I agree with @t0yv0 that it's a little confusing at the moment to have this in the upgrade provider command since it doesn't actually perform an upgrade. But I'm also OK with shipping and iterating since this is an internal tool.

main.go
guineveresaenger: I hear you on this. My personal end goal is to remove issue creation altogether but if that turns out to be too ambitious, we can add extra flags.

PR #212, opened 2023-12-08
Detect and upgrade Java Gen Version.

guineveresaenger:
This PR depends on a few provider repository structure tweaks as prerequisites.

1. In order to properly track and check in the current Java Gen version, we will write this version to a file `.pulumi/java-gen-version`. Native providers use this pattern as a dependencies tracker already so this is not new.
2. We will start to check in the version files in the `.pulumi` folder - both `version` (meaning pulumi itself) and `java-gen-version`. The plugins will remain gitignored.
3. In the Makefile, rather than reading in the `JAVA_GEN_VERSION` during a CI upgrade (which does not build SDKs), we will depend on `.pulumi/java-gen-version` for the `bin/pulumi-java-gen` target. [Example Makefile changes](https://github.com/pulumi/pulumi-aiven/pull/446/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)(https://github.com/pulumi/pulumi-aiven/pull/446/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)
4. _Only a full upgrade using upgrade-provider (or running the equivalent manual steps)_ will be updating the `.pulumi/java-gen-version` file.
5. We slightly change the meaning of `javaGenVersion` for ci and upgrade configs; we will use this option as a means to pin the Java Gen version, should we need to ever roll back in a provider.

This PR lays down the groundwork for points  1, 4, and 5.

- add new boolean flag `--upgrade-java` that defaults to true
- check the version read from `.pulumi/pulumi-java-gen`, compare to the latest version of pulumi-java, and upgrade when necessary
- we retain the ability to pin a version via `upgrade-provider provider/repo --java-version=x.y.z`. In this case, we set the `.pulumi/java-gen-version` to that version, and do not upgrade to anything newer.

Parts 2&3 will be changed via ci-mgmt.

Part of https://github.com/pulumi/ci-mgmt/issues/506.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pulumi/upgrade-provider/colorize"
 	"github.com/pulumi/upgrade-provider/step"
 	stepv2 "github.com/pulumi/upgrade-provider/step/v2"
+	"github.com/ryboe/q"
 )
 
 // A "git commit" step that is resilient to no changes in the directory.
@@ -157,6 +158,9 @@ func UpgradeProviderVersion(
 	repo ProviderRepo, targetSHA, forkedProviderUpstreamCommit string,
 ) step.Step {
 	steps := []step.Step{}
+	q.Q(GetContext(ctx))
+	q.Q(GetContext(ctx).JavaVersion)
+
 	if javaVersion := GetContext(ctx).JavaVersion; javaVersion != "" {
 		var didChange bool
 		steps = append(steps, step.Combined("Update Java Version",
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -328,6 +328,8 @@ var InformGitHub = stepv2.Func70E("Inform Github", func(
 		prTitle = "Upgrade pulumi-terraform-bridge/pf to " + targetPfVersion.String()
 	case c.TargetPulumiVersion != nil:
 		prTitle = "Test: Upgrade pulumi/{pkg,sdk} to " + c.TargetPulumiVersion.String()
+	case c.UpgradeJavaVersion:
+		prTitle = "Upgrade pulumi-java to" + c.JavaVersion
 	default:
 		return fmt.Errorf("Unknown action")
 	}
@@ -529,6 +531,8 @@ var getWorkingBranch = stepv2.Func41E("Working Branch Name", func(ctx context.Co
 		return ret("upgrade-pf-version-to-%s", targetPfVersion)
 	case c.TargetPulumiVersion != nil:
 		return ret("upgrade-pulumi-version-to-%s", c.TargetPulumiVersion)
+	case c.UpgradeJavaVersion:
+		return ret("upgrade-java-version-to-%s", c.JavaVersion)
 	default:
 		return "", fmt.Errorf("calculating branch name: unknown action")
 	}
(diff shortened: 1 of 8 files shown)


t0yv0:
Too bad `pulumi-language-java --version` does not work as its' deployed right alongside pulumi CLI and we could just get it from there.

t0yv0:
Something I would like help understanding here is why use this path:

```
.pulumi/java-gen-version
```

I believe this path will be under `.gitignore`? So in that case any changes that upgrade-provider makes will not be persisted?

Why not a top-level path like this:

https://github.com/pulumi/pulumi-azure-native/blob/master/.pulumi-java-gen.version

t0yv0:
https://github.com/pulumi/pulumi-aws/blob/master/.gitignore#L4 the presence of this kind of .gitignore entry in the wild makes the path choice very unfortunate.

upgrade/util.go
```diff
@@ -38,6 +38,8 @@ type Context struct {
 	UpgradeProviderVersion bool
 	MajorVersionBump       bool
 
+	UpgradeJavaVersion bool
```
t0yv0, line 41: Why do we have to opt into it? Why not upgrade it every time we update codegen?

guineveresaenger:
I am planning to remove `.pulumi` from the `gitignore` via source migration. I wanted to follow the existing pattern we have for where we read from versions, since we already read from this place for the pulumi version.
The plugin files in `.pulumi/bin` would continue to be ignored.

upgrade/util.go
guineveresaenger: It currently defaults to true so in effect it does upgrade every time unless explicitly set to false. I also imagined it would give us a more flexible safeguard in case we ever have a bad pulumi-java version and want to halt upgrades.

But you're right that we can likely do without.

t0yv0: Ahh that's nice. That's great then.

t0yv0:
>  since we already read from this place for the pulumi version.

That is specifically done intentionally so that it would be .gitignore'd.

guineveresaenger:
> That is specifically done intentionally so that it would be .gitignore'd.
I'm probably missing some context here - is it strictly necessary that the pulumi version be gitignored? I suppose it is really just a byproduct of reading from go.mod and getting plugins, so that makes sense.

> That can work but isn't it much simpler to pick a path that's not ignored in the first place?
Sure - I can do that too; I have no real strong feelings here!

upgrade/steps.go
```diff
@@ -150,47 +150,11 @@ var ensureUpstreamRepo = stepv2.Func11("Ensure Upstream Repo", func(ctx context.
...
 })
 
-var javaVersionRegexp *regexp.Regexp = regexp.MustCompile(`JAVA_GEN_VERSION := (v[0-9]+\.[0-9]+\.[0-9]+)`)
-
 func UpgradeProviderVersion(
 	ctx context.Context, goMod *GoMod, target *semver.Version,
 	repo ProviderRepo, targetSHA, forkedProviderUpstreamCommit string,
 ) step.Step {
 	steps := []step.Step{}
-	if javaVersion := GetContext(ctx).JavaVersion; javaVersion != "" {
```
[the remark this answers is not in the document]
guineveresaenger: This code only updates `JAVA_GEN_VERSION` which the parallel ci-mgmt changes will remove the need for.

PR #214, opened 2023-12-13
Correctly obtain upstream org name for provider

### LINKED ISSUE #213: `go.Mod.UpstreamProviderOrg` is unreliable and misleading
### What happened?

For the new [upstream upgrade check](https://github.com/pulumi/pulumi-gitlab/blob/master/.github/workflows/check-upstream-upgrade.yml) we use `go.Mod.UpstreamProviderOrg` to detect upstream releases.

However, the way that field gets populated is by attempting to read the upstream provider from the `provider/go.mod` file, which:

- returns the first found match for `*/terraform-provider-foo` 
- does not account for submodules that have since been reamed (such as [DataDog](https://github.com/pulumi/pulumi-datadog/blob/master/provider/go.mod#L17) - the org is `DataDog` not `terraform-providers`.
- does not account for gitlab having a [mirror repository](https://github.com/gitlabhq/terraform-provider-gitlab), which [never makes it into our go.mod](https://github.com/pulumi/pulumi-gitlab/blob/master/provider/go.mod#L18)

- in summary, may or may not be completely different from the relevant upstream org/repo that we would want to check versions from.

This behavior is correct for "typical" providers with no renames, replaces, or otherwise obscuring the link to the GitHub org/repo location of the upstream provider, so it "works" in those instances.

The reason we did not detect this bug sooner is that we did not ever actually use the `go.Mod.UpstreamProviderOrg` before attempting to discover upstream upgrades.

### Example

https://github.com/pulumi/pulumi-gitlab/issues/445

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

guineveresaenger:
Inferring the upstream provider org from the go mod requirements is not as robust as once believed.
This pull request reads from provider/resources.go instead, looking for the value of `GitHubOrg`.

This change will unblock Stateless Provider Upgrades rollout, and it will additionally ensure that the fallback behavior for the "Get Expected Target / From Upstream Releases" works correctly for all providers.

A successful run using this commit is here: https://github.com/pulumi/pulumi-gitlab/actions/runs/7201244400/job/19616931040 - it is broken using `upgrade-provider@main`.

Fixes #213

### DIFF
--- a/upgrade/kind.go
+++ b/upgrade/kind.go
@@ -186,7 +186,7 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
 		Pf:       pf,
 	}
 
-	upstreamOrg := stepv2.Func11("Get UpstreamOrg", func(ctx context.Context, path string) string {
+	out.UpstreamProviderOrg = stepv2.Func11("Get UpstreamOrg", func(ctx context.Context, path string) string {
 		stepv2.SetLabel(ctx, "From resources.go")
 		resourceFile := filepath.Join(path, "provider", "resources.go")
 		resourceData := stepv2.ReadFile(ctx, resourceFile)
@@ -207,10 +207,6 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
 		}
 		return upstreamOrg
 	})(ctx, path)
-
-	out.UpstreamProviderOrg = upstreamOrg
-
-	// get the org name that hosts the upstream repo from the GitHubOrg field in provider/resources.go
 
 	if fork == nil {
 		out.Kind = Plain
--- a/upgrade/kind.go
+++ b/upgrade/kind.go
@@ -185,9 +185,22 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
 		Bridge:   bridge,
 		Pf:       pf,
 	}
-	// get the org name that hosts the upstream repo
-	tok := strings.Split(modPathWithoutVersion(upstream.Mod.Path), "/")
-	out.UpstreamProviderOrg = tok[len(tok)-2]
+	// get the org name that hosts the upstream repo from the GitHubOrg field in provider/resources.go
+	resourceFile := filepath.Join(path, "provider", "resources.go")
+	resourceData := stepv2.ReadFile(ctx, resourceFile)
+	resLines := strings.Split(resourceData, "\n")
+	var upstreamOrg string
+	for _, line := range resLines {
+		if strings.Contains(line, "GitHubOrg") {
+			lineSlice := strings.Split(line, "\"")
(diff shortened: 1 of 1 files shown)


upgrade/kind.go
```diff
@@ -185,9 +185,22 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
...
+	var upstreamOrg string
+	for _, line := range resLines {
+		if strings.Contains(line, "GitHubOrg") {
+			lineSlice := strings.Split(line, "\"")
+			upstreamOrg = lineSlice[1]
+			break
+		}
+	}
+	if upstreamOrg == "" {
+		contract.Assertf(upstreamOrg != "nil", "upstreamOrg cannot be empty; set GitHubOrg in ProviderInfo")
```
t0yv0, line 201: Eh,  this really not robust at all and will cause problems, reading source code .. there's no guarantee it's a literal or anything. I know this can be a pain to extract but we need to file a ticket to get the right. Also if we're going ahead with these approximate methods how about falling back gently? Fall back on the previously working method if not detected robustly by this new method? 


This field is explicitly optional in the bridge.

```
func (info ProviderInfo) GetGitHubOrg() string {
	if info.GitHubOrg == "" {
		return "terraform-providers"
	}

	return info.GitHubOrg
}

```

t0yv0: The "right"-er way would be either this moves to a config file or if it's in Go source then we compile the go file and call the Go software to self-describe.

t0yv0 (changes requested):
I'd approve modulo one blocker: please make sure this does not hard-fail if the optional field GitHubOrg is either not set or fails to parse. There's a slight chance upgrade-provider is used outside the organization and it may arise that somewhere this field is not set, and this PR makes it required.

upgrade/kind.go
guineveresaenger: Yeah, the options are not great.

We can of course fall back on the previous implementation if we cannot find an upstream org scanning the resources.go file.

But that doesn't solve for the case where `GitHubOrg` is set incorrectly (nor would compiling the source code and querying ProviderInfo). 
Submodule detection does not work either, for the same reason this is a bug in the first place.

These are all reasonable arguments for using an upgrade config but I was hoping we could avoid manual input. :/

upgrade/kind.go
```diff
@@ -185,9 +185,24 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
...
+	var upstreamOrg string
+	for _, line := range resLines {
+		if strings.Contains(line, "GitHubOrg") {
+			lineSlice := strings.Split(line, "\"")
+			upstreamOrg = lineSlice[1]
+			break
+		}
+	}
+	if upstreamOrg == "" {
+		// fall back to attempting to read from go.mod
```
t0yv0, line 201: Can log this.

t0yv0 (approved):

PR #215, opened 2023-12-14
Check file exists before reading upstream org

guineveresaenger:
Prevent panics such as https://github.com/pulumi/pulumi-azure/issues/1514.


guineveresaenger:
this is deeply frustrating; locally the linter isn't throwing these :/

PR #219, opened 2023-12-15
Convert "Planning Bridge Upgrade" to stepv2

iwahbe:
This should be a no-op from the user perspective. It does enable us to test this step.


step/v2/step.go
```diff
@@ -131,6 +134,13 @@ func SetLabel(ctx context.Context, label string) {
 	p.handleError([]any{err})
 }
 
+// SetLabel displays the formatted label next to the currently running step.
```
guineveresaenger, line 137:
```suggestion
// SetLabelf displays the formatted label next to the currently running step.
```

step/v2/step.go
```diff
@@ -243,3 +253,23 @@ func (p *pipeline) errExit(err error) {
...
+
+// cast performs a type cast from src to T.
+//
+// Unlike src.(T), this cast is valid when casting from an untyped nil to a typed nil.
+func cast[T any](src any) T {
+	if src != nil {
+		return src.(T)
+	}
+	var t T
+	typ := reflect.TypeOf(t)
```
guineveresaenger, line 265: I'm curious as to the specific use case - I assume a very particular nil check after the function returns?

upgrade/steps_helpers.go
```diff
@@ -347,16 +346,18 @@ func setUpstreamFromRemoteRepo(
...
-		return gitRepoRefs{}, fmt.Errorf("git %s: %w", strings.Join(args, " "), err)
-	}
+func gitRefsOf(ctx context.Context, url, kind string) (refs gitRepoRefs, err error) {
+	err = stepv2.PipelineCtx(ctx, "shim", func(ctx context.Context) {
+		refs = gitRefsOfV2(ctx, url, kind)
+	})
+	return
+}
+
+var gitRefsOfV2 = stepv2.Func21("git refs of", func(ctx context.Context, url, kind string) gitRepoRefs {
```
guineveresaenger, line 356: Is there a reason we want this to be a var? Do we need to mutate this later?

upgrade/steps_helpers.go
```diff
@@ -404,24 +404,34 @@ func (g gitRepoRefs) sortedLabels(less func(string, string) bool) []string {
...
-		repo, "--json=latestRelease").Output()
-	if err != nil {
-		return nil, err
-	}
+var findCurrentMajorVersion = stepv2.Func21("Find Current Major Version",
+	func(ctx context.Context, repoOrg, repoName string) *semver.Version {
+		repoCurrentVersion := latestRelease(ctx, repoOrg+"/"+repoName)
+		stepv2.SetLabelf(ctx, "%d", repoCurrentVersion.Major())
+		return repoCurrentVersion
+	})
```
guineveresaenger, line 412: complete nit because I have a weird brain: can we do this for readability?

```suggestion
	},
)
```

upgrade/upgrade_provider.go
```diff
@@ -176,61 +174,60 @@ func UpgradeProvider(ctx context.Context, repoOrg, repoName string) (err error)
 				latestSha, ok := refs.shaOf(latestTag)
 				contract.Assertf(ok, "Failed to lookup sha of known tag: %q not in %#v",
 					latestTag, refs.labelToRef)
+
+				// Assign out of planning function here.
```
guineveresaenger, line 178: What is "out of planning" in this case?

step/v2/step.go
```diff
@@ -243,3 +253,23 @@ func (p *pipeline) errExit(err error) {
 	p.failed = err
 	runtime.Goexit()
 }
+
+// cast performs a type cast from src to T.
+//
+// Unlike src.(T), this cast is valid when casting from an untyped nil to a typed nil.
+func cast[T any](src any) T {
```
guineveresaenger, line 260: `cast` is a very overloaded word!

Let's give this function a more descriptive name, like `castWithTypedNil` or something like that. I know it's a library function, and I know we use it literally hundreds of times but we can be a bit more specific I think!

upgrade/steps_helpers.go
```diff
@@ -404,24 +404,34 @@ func (g gitRepoRefs) sortedLabels(less func(string, string) bool) []string {
...
+	err := json.Unmarshal([]byte(resultString), &result)
 	if err != nil {
 		return nil, err
 	}
 
-	return semver.NewVersion(result.Latest.TagName)
-}
+	v, err := semver.NewVersion(result.Latest.TagName)
+	if err == nil {
+		stepv2.SetLabelf(ctx, "of %s: %s", repo, v)
```
guineveresaenger, line 430: yes I like this! makes the return more explicit too.

guineveresaenger (approved):
I had a few questions and a few nits but let's get this merged.

Is there a sample upgrade run using this version?

step/v2/step.go
iwahbe: I simplified that function. Basically, `maybeNil.(T)` panics when `maybeNil` is in fact `nil`.

upgrade/steps_helpers.go
iwahbe: We do not want to mutate it later. It's a `var` because it can't be a `const`.

guineveresaenger: Can it not be a `func`?

upgrade/upgrade_provider.go
iwahbe: Cleaned up.

upgrade/steps_helpers.go
iwahbe: Not without pain, because we want it to be wrapped in `stepv2.Func21("git refs of", ...)`. The equivalent `func` would be this:

```go
func gitRefsOfV2(ctx context.Context, url, kind string) gitRepoRefs {
	return stepv2.Func21("git refs of", func(ctx context.Context, url, kind string) gitRepoRefs {
		out := stepv2.Cmd(ctx, "git", "ls-remote", "--"+kind, url)

		branchesToRefs := map[string]string{}
		for i, line := range strings.Split(out, "\n") {
			if line == "" {
				continue
			}
			parts := strings.Split(line, "\t")
			contract.Assertf(len(parts) == 2,
				"expected git ls-remote to give '\t' separated values, found line %d: '%s'",
				i, line)
			branchesToRefs[parts[1]] = parts[0]
		}
		return gitRepoRefs{branchesToRefs, kind}
	})(ctx, url, kind)
}
```

That's a lot of code duplication and extra indentation for enforced immutability.

step/v2/step.go
guineveresaenger: this is still missing the `f`... 😅

upgrade/steps_helpers.go
guineveresaenger: I have differing preferences here but hey're 100% non-blocking. It does look annoying to write; let's go with the var implementation.

step/v2/step.go
iwahbe: I don't want to give the function a name like `castWithTypedNil`, since the point of the function is to hide the need to treat `nil`s differently from non-`nil` values. When we discover more problems with this function (and I'm sure we will), I don't want to change the function name. I could call it `assertType[T any](src any) T`, since go calls `src.(T)` a type assertion, but I'm not sure if that adds any clarity here.

Do you think that `assertType` is better?

PR #220, opened 2023-12-16
Use semver sort to determine latest upstream release in getExpectedTargetLatest

### LINKED ISSUE #218: Failing to detect a new major version
### What happened?

We ran this workflow: https://github.com/pulumi/pulumi-minio/actions/runs/7225716362/job/19689733780, reporting no new upgrade targets:
Screenshot 2023-12-15 at 11 01 36 AM

However, we were on a `v1` version, while upstream was on a `v2` version.

### Example

NA

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

guineveresaenger:
See #218 for details.

Uses the full json repo info payload to determine the latest upstream release.

See https://github.com/pulumi/pulumi-minio/issues/251 for the fix applying to pulumi-minio specifically.

Fixes #218.
Fixes https://github.com/pulumi/pulumi-f5bigip/issues/345
Fixes https://github.com/pulumi/pulumi-venafi/issues/244

~**Reviewer Note**: This functionality overrides the changes in #226. Note that we're returning an empty UpgradeTargetVersion in the case of a non-stable release, so that the nil check on `Plan Provider Upgrade` reports that we are up to date. A `latest` version with a beta tag should not fail the upgrade process; we should ignore it.~

### DIFF
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -505,23 +505,69 @@ var getExpectedTarget = stepv2.Func21("Get Expected Target", func(ctx context.Co
 	return getExpectedTargetLatest(ctx, name, upstreamOrg)
 })
 
-// Figure out what version of upstream to target by searching for the latest GitHub
-// release.
+// getExpectedTargetLatest discovers the latest stable release and sets it on UpstreamUpgradeTarget.Version.
+// There is a lot of human error and differing conventions when discovering and defining the "latest" upstream version.
+// For our purposes, we always want to discover the highest, stable, valid Semver version of the upstream provider.
+// We do so by listing the last 30 GitHub releases, extracting the tags from the output result (we eagerly await being
+// able to get this result in json), parsing the tags into versions (filtering out any invalid or non-stable tags),
+// and sorting them.
 var getExpectedTargetLatest = stepv2.Func21E("From Upstream Releases", func(ctx context.Context,
 	name, upstreamOrg string) (*UpstreamUpgradeTarget, error) {
 
 	upstreamRepo := upstreamOrg + "/" + GetContext(ctx).UpstreamProviderName
+	// TODO: use --json once https://github.com/cli/cli/issues/4572 is fixed
+	releases := stepv2.Cmd(ctx, "gh", "release", "list",
+		"--repo="+upstreamRepo,
+		"--exclude-drafts",
+		"--exclude-pre-releases")
 
-	version, err := latestRelease(ctx, upstreamRepo)
-	if err != nil {
-		return nil, err
+	resultLines := strings.Split(releases, "\n")
+	// Get version tags. This will become much less laborious once we can use json.
+	var tags []string
+	for _, line := range resultLines {
+		// split the result line by whitespace
+		fields := strings.Fields(line)
+		// handle empty newlines
+		if len(fields) == 0 {
+			continue
+		}
+		// the tag name for the release is the second field of the result line.
+		tag := fields[1]
+		// except when the release is marked as Latest, then it's the third field.
+		if tag == "Latest" {
+			tag = fields[2]
+		}
+		tags = append(tags, tag)
 	}
-	// In some cases, upstream marks non-stable releases as `latest`.
-	// Because we do not upgrade on those, we return an empty upstream target.
-	if version.Prerelease() != "" {
-		return &UpstreamUpgradeTarget{}, nil
+
+	// Parse tags into versions
+	var versions []*semver.Version
+
+	for _, tag := range tags {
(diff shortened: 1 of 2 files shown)


VenelinMartinov:
Thanks for fixing the issue here! I'm not sure this is quite the right logic though. It looks to me like this change yields the following logic:
If latest is a beta, do not upgrade.  
This is not the correct behaviour in the following example:  
1.2-beta latest
1.1
1.0 (current)

We would not upgrade to 1.1 as we should.

Should we instead get the latest version by ordering the list from gh releases?

guineveresaenger:
I mean, this is all trying to make up for inconsistent human behavior around setting release standards.

Do we think the `latest` means anything? It sounds like we do not trust upstream to use GitHub `latest` consistently, as evidenced by a non-stable version being marked as `latest` on the postgresql provider.

However, I think the case you describe is unlikely:
- [Terraform requires you to publish a github release](https://developer.hashicorp.com/terraform/registry/providers/publishing#creating-a-github-release)
- GitHub pre-selects "Set As Latest Release" when you create a release in the UI, and release tools such as we use do the same. 

Likely, what happened here is that a maintainer failed to check "Set as pre-release" when they published their beta version. I would be surprised to see a stable release not setting github `latest` by default. (It's not impossible).

The reason why I prefer checking things in this way is that there have been several issues in this tool with hand ordering tags, and asking for "what is latest on github" is pretty foolproof from an implementation perspective.

I would argue it is reasonable to assume `latest` is good, and guard for accidental latest-has-a-prerelease-semver by ignoring it.

guineveresaenger:
Alright I spoke with @iwahbe about this.

Because (github)`latest` moves and we only check for updates once a day, we may in fact miss "latest" versions due to `latest` moving on to a newer tag, and if that tag is not stable semver, we do not upgrade.

However, we also cannot use the first release from `gh release list` because those are ordered by publish date, not by semver.

We will do a best-attempt semver comparison of the default result of `gh release list` (30 most recent releases).

...I do not like this. But it is the least lossy option we have. :(

guineveresaenger:
An upgrade issue with the correct upstream version was [opened successfully using this code](https://github.com/pulumi/pulumi-minio/issues/264).

guineveresaenger:
This should also be fixing several of the failures from https://github.com/pulumi/upgrade-provider/issues/225.

guineveresaenger:
Hm, the lack of json output is _extremely_ unfortunate. Adjusting to split lines by tab instead to ensure we don't wind up with  "Terraform" as our release tag when there is whitespace in release names. This should fix [this failure](https://github.com/pulumi/pulumi-f5bigip/actions/runs/7217436388/job/19665253187).

```
gh release list --repo=F5Networks/terraform-provider-bigip --exclude-drafts --exclude-pre-releases
TITLE                              TYPE    TAG NAME  PUBLISHED
Terraform Provider BIG-IP v1.20.1  Latest  v1.20.1   about 1 month ago
Terraform Provider BIG-IP v1.20.0          v1.20.0   about 2 months ago
Terraform Provider BIG-IP v1.19.0          v1.19.0   about 4 months ago
Terraform Provider BIG-IP v1.18.1          v1.18.1   about 5 months ago
Terraform Provider BIG-IP v1.18.0          v1.18.0   about 6 months ago
Terraform Provider BIG-IP v1.17.1          v1.17.1   about 8 months ago
Terraform Provider BIG-IP v1.17.0          v1.17.0   about 9 months ago
Terraform Provider BIG-IP v1.16.2          v1.16.2   about 11 months ago
Terraform Provider BIG-IP v1.16.1          v1.16.1   about 1 year ago
Terraform Provider BIG-IP v1.16.0          v1.16.0   about 1 year ago
Terraform Provider BIG-IP v1.15.2          v1.15.2   about 1 year ago
```

upgrade/steps_helpers.go
```diff
@@ -502,35 +502,69 @@ var getExpectedTarget = stepv2.Func21("Get Expected Target", func(ctx context.Co
...
+	}
+	// if we did not find any valid versions, we return.
+	if len(versions) == 0 {
+		return nil, fmt.Errorf("no valid stable versions found in %s", upstreamRepo)
 	}
-	return nil, fmt.Errorf("no non-beta releases found in %s/%s",
-	upstreamOrg, GetContext(ctx).UpstreamProviderName)
+	// Sort the versions.
+	// Documentation here: https://pkg.go.dev/github.com/Masterminds/semver/v3#readme-sorting-semantic-versions
+	sort.Sort(semver.Collection(versions))
```
VenelinMartinov, line 562: 🔥

VenelinMartinov (approved):

iwahbe (approved):
LGTM

PR #226, opened 2023-12-27
Ignore beta releases

VenelinMartinov:
Fixes https://github.com/pulumi/ci-mgmt/issues/749

When looking up the latest version of an upstream provider, we should skip beta releases.  
This PR achieves this by getting the list of the latest 30 releases and then iterating over them until it finds a non-beta release when looking up latest.

### DIFF
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


upgrade/steps_helpers.go
```diff
@@ -511,18 +511,25 @@ var getExpectedTargetLatest = stepv2.Func21E("From Upstream Releases", func(ctx
 	name, upstreamOrg string) (*UpstreamUpgradeTarget, error) {
 	latest := stepv2.Cmd(ctx, "gh", "release", "list",
```
iwahbe, line 512: We should use `--json` when possible. Tracking issue is https://github.com/cli/cli/issues/4572.

upgrade/steps_helpers.go
```diff
@@ -511,18 +511,25 @@ var getExpectedTargetLatest = stepv2.Func21E("From Upstream Releases", func(ctx
...
-		upstreamOrg, GetContext(ctx).UpstreamProviderName))
-	v, err := semver.NewVersion(tok[0])
-	if err != nil {
-		return nil, err
+	
+	versions := strings.Split(latest, "\n")
+	for _, ver := range versions {
+		tok := strings.Fields(ver)
+		contract.Assertf(len(tok) > 0, fmt.Sprintf("no releases found in %s/%s",
+			upstreamOrg, GetContext(ctx).UpstreamProviderName))
```
iwahbe, line 521: The `f` in `contract.Assertf` means that it acts like `fmt.Sprintf` for its inputs. This is a pretty common go idiom.
```suggestion
		contract.Assertf(len(tok) > 0, "no releases found in %s/%s",
			upstreamOrg, GetContext(ctx).UpstreamProviderName)
```

upgrade/steps_helpers_test.go
```diff
@@ -292,3 +290,50 @@ func TestGetExpectedTargetFromTarget(t *testing.T) {
   "GHIssues": null
 }`))
 }
+
+func TestFromUpstreamReleasesBetaIgnored(t *testing.T) {
```
iwahbe, line 294: Thanks for adding a test here!

iwahbe (approved):
One nit, but otherwise LGTM.

PR #228, opened 2024-01-03
Use upstream org config

### LINKED ISSUE #221: Need for upgrade-specific `upstreamOrg` config
`check-upstream-upgrade` [failed for pulumi-dnsimple](https://github.com/pulumi/pulumi-dnsimple/actions/runs/7217418099) because it could not find `terraform-providers/terraform-provider-dnsimple`. This is because the terraform provider has since moved to `github.com/dnsimple/terraform-provider-dnsimple`.

When attempting to [add `GitHubOrg: dnsimple` to the provider info in `resources.go`](https://github.com/pulumi/upgrade-provider/pull/214) and regenerated the documentation, I discovered that _all the docs were being dropped_ because, while the github repo name is `dnsimple/terraform-provider-dnsimple`, the _package name_ remains `terraform-providers/terraform-provider-dnsimple`, and `GitHubOrg` is being used to generate documentation via go mod download. 😃 

_Removing_ `GitHubOrg` also does nothing for detecting the upstream version correctly, because the fallback reads from go.mod's Requires, which is _also_ `terraform-providers/terraform-provider-dnsimple`.

Therefore, asking `GitHubOrg` to fulfill the secondary duty of upstream org discovery is unfortunately not failproof. We need an upstreamOrg field in the upgrade config, and use that for the specific purpose of detecting upstream versions.

The nice thing here is that when repos get renamed and moved to a different org, we will find out immediately.

guineveresaenger:
This PR reverts part of the changes made in #214 in favor of using an explicit new config var, `upstreamProviderOrg`. 
The use context here focuses on having a _valid git host url_ for detecting new versions upstream that is separate from the _go module path_.

When this config variable is set on the provider's `.ci-mgmt` config file, `upgrade-provider` will read this value first and set it in the context.
When it is unset, we will continue to infer the upstream org from the go.mod file as before.
We are removing parsing the resources.go file for the `GitHubOrg` value as that use case requires a valid go module path.

The rollout here will cause a few providers to fail their nightly upstream version check. The fix is to add the correct `upstreamProviderOrg` to the CI config.
[Corresponding ci-mgmt PR](https://github.com/pulumi/ci-mgmt/pull/770).

Additionally, this PR moves some org mapping configuration out of this tool to be configured on provider level instead. In order to keep logic straightforward, one option is to hard code the GitLab provider's remote - if that is objectionable, please only consider changes up to https://github.com/pulumi/upgrade-provider/pull/228/commits/c22dec245d9c706ae6f35e0e88ea6b259f2358e9.

Fixes #221.

- Read in upstream provider org from upgrade config
- We do not need to call getGitHubPath for the pulumi provider path; it is already normalized.
- Use context to substitute the ustream org instead of ProviderOrgs
- Remove owner/repo mappings


upgrade/kind.go
```diff
@@ -192,28 +192,19 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
 	}
 
 	out.UpstreamProviderOrg = stepv2.Func11("Get UpstreamOrg", func(ctx context.Context, path string) string {
-		resourceFile := filepath.Join(path, "provider", "resources.go")
-		_, found := stepv2.Stat(ctx, resourceFile)
+		//resourceFile := filepath.Join(path, "provider", "resources.go")
+		//_, found := stepv2.Stat(ctx, resourceFile)
```
iwahbe, line 196: Let's delete the comment here.
```suggestion
```

upgrade/kind.go
```diff
@@ -192,28 +192,19 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
...
+		upstreamOrg = GetContext(ctx).UpstreamProviderOrg
+		if upstreamOrg != "" {
+			stepv2.SetLabel(ctx, "From Upgrade Config")
+			return upstreamOrg
 		}
+		// fall back to attempting to read from go.mod
+		stepv2.SetLabel(ctx, "Fall back to reading from go.mod")
+		tok := strings.Split(modPathWithoutVersion(upstream.Mod.Path), "/")
+		upstreamOrg = tok[len(tok)-2]
+
```
iwahbe, line 207: The logic is much simpler 🎉! Code wise, I think this could be made simpler still.

In general, I think that we should avoid code like this:

```go
		var upstreamOrg string
		upstreamOrg = GetContext(ctx).UpstreamProviderOrg
		// break
		upstreamOrg = tok[len(tok)-2]
		return upstreamOrg
```

Instead, we should write:

```go
		upstreamOrg := GetContext(ctx).UpstreamProviderOrg
		// break
		// We expect tokens to be of the form:
		//
		//	github.com/${org}/${repo}/${path}
		//
		// The second chunk is the org name.
		return tok[len(tok)-2]
```

☝️ actually calls `tok[len(tok)-2]` into question. Most TF providers have a top level module (path = `github.com/${org}/${repo}`), but go allows nested modules (path = `github.com/${org}/${repo}/${nest-path}`). Since we are assuming GitHub, `tok[len(tok)-2]` will fail on nested paths, while `tok[1]` will always work.

```suggestion
		if org := GetContext(ctx).UpstreamProviderOrg; org != "" {
			stepv2.SetLabelf(ctx, "From Upgrade Config: %s", org)
			return org
		}
		// fall back to attempting to read from go.mod
		stepv2.SetLabel(ctx, "Fall back to reading from go.mod")
		tok := strings.Split(modPathWithoutVersion(upstream.Mod.Path), "/")
		return tok[1]

```

upgrade/provider_owners.go
iwahbe, line 1: 🎉🔥🎉🔥🎉

upgrade/steps.go
```diff
@@ -182,14 +179,19 @@ func UpgradeProviderVersion(
...
+				gitHostPath := "https://github.com/" + upstreamOrg + "/" + upstreamRepo
+
+				// special case: we need to use the GitLab url for getting git refs.
+				if upstreamOrg == "terraform-provider-gitlab" {
+					gitHostPath = "https://gitlab.com/gitlab-org/terraform-provider-gitlab"
 				}
-				refs, err := gitRefsOf(ctx, "https://"+modPathWithoutVersion(path),
+
+				refs, err := gitRefsOf(ctx, gitHostPath,
 					"tags")
```
iwahbe, line 195: The line is now pretty short (within lint bounds), so let's put ` "tags")` back on the same line.
```suggestion
				refs, err := gitRefsOf(ctx, gitHostPath, "tags")
```

upgrade/kind.go
```diff
@@ -192,28 +192,19 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
 	}
 
 	out.UpstreamProviderOrg = stepv2.Func11("Get UpstreamOrg", func(ctx context.Context, path string) string {
```
iwahbe, line 194: It would be great to separate out `stepv2.Func11("Get UpstreamOrg", ...` into a standalone function (`var getUpstreamOrg = stepv2.Func11("Get UpstreamOrg", ...`) and put it under test.

Also, `path` is no longer used, so we can remove it.

upgrade/util.go
```diff
@@ -47,6 +47,7 @@ type Context struct {
 	//	pulumi-aws
 	//
 	UpstreamProviderName string
+	UpstreamProviderOrg  string
```
iwahbe, line 50: This was confusing, so let's document it.

```suggestion
	// The org component in the upstream provider's repo path.
	// 
	// This must be provided when upstream is hosted at a repo that does 
	// not match the Go Module path it is declared as. Otherwise it can 
	// be inferred.
	//
	// For example, if we have an upstream provider with import path:
	//
	//	github.com/terraform-providers/terraform-provider-my-provider
	//
	// which is hosted at:
	//
	//	github.com/my-org/terraform-provider-my-provider
	//
	// Then UpstreamProviderOrg should be `my-org`.
	UpstreamProviderOrg  string
```

iwahbe (commented):
The new logic looks good. I left a couple comments to address before we merge.

upgrade/kind.go
guineveresaenger: that's a good catch!

upgrade/kind.go
guineveresaenger: Done

iwahbe (approved):
LGTM!

Thanks for the excellent tests.

PR #229, opened 2024-01-06
Enable 8-week release cadence

guineveresaenger:
Prior to this change, all provider releases via this tool only occurred on upstream upgrades. This meant that security and codegen/bridge updates would merge to the default branch, but wait on an upstream release to apply to the provider, potentially resulting in a provider using an ancient version of the bridge.

This pull request sets a maximum ~8-week release cadence for all providers.

During a bridge update, we check if we have released the provider in question within the last 8 weeks.
If we have not, and the upgrade is a non-upstream upgrade, we will add a `needs-release/patch` label to the bridge update pull request.

Doing this will result in more frequent upgrades, ensuring our providers use a recent bridge and pulumi version.

One detail to call out:
- We may choose to set the release cadence to 7 weeks, to ensure we "catch" the update at minimum every 8th week; this PR sets it to 8 weeks exactly.
- We can also count actual bridge releases if we do not want to use time comparisons.

Fixes https://github.com/pulumi/ci-mgmt/issues/726.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -267,7 +267,7 @@ var maintenanceRelease = stepv2.Func11E("Check if we should release a maintenanc
 	ctx context.Context,
 	repo ProviderRepo,
 ) (bool, error) {
-	repoWithOrg := repo.org + "/" + repo.name
+	repoWithOrg := repo.Org + "/" + repo.Name
 	// We ensure a release at least every 8-9 weeks, concurrent with a bridge update.
 	// There are 24 * 7 * 8 = 1344 hours in 8 weeks.
 	releaseCadence := time.Hour * 24 * 7 * 8
@@ -846,7 +846,7 @@ func migrationSteps(ctx context.Context, repo ProviderRepo, providerName string,
 }
 
 func AddAutoAliasing(ctx context.Context, repo ProviderRepo) (step.Step, error) {
-	providerName := strings.TrimPrefix(repo.name, "pulumi-")
+	providerName := strings.TrimPrefix(repo.Name, "pulumi-")
 	metadataPath := fmt.Sprintf("%s/cmd/pulumi-resource-%s/bridge-metadata.json", *repo.providerDir(), providerName)
 	steps := []step.Step{
 		step.F("ensure bridge-metadata.json", func(context.Context) (string, error) {
@@ -871,7 +871,7 @@ func AddAutoAliasing(ctx context.Context, repo ProviderRepo) (step.Step, error) 
 }
 
 func ReplaceAssertNoError(ctx context.Context, repo ProviderRepo) (step.Step, error) {
-	steps, err := migrationSteps(ctx, repo, repo.name, "Remove deprecated contract.Assert",
+	steps, err := migrationSteps(ctx, repo, repo.Name, "Remove deprecated contract.Assert",
 		AssertNoErrorMigration)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


upgrade/steps.go
```diff
@@ -315,6 +349,12 @@ var InformGitHub = stepv2.Func70E("Inform Github", func(
 			}
 		}
 
+		// On non-upstream upgrades, we will create a patch release label
+		// if the provider hasn't been released in 8 weeks.
+		if c.MaintenancePatch && !c.UpgradeProviderVersion {
```
iwahbe, line 354: Optional nit: I would put these in a `switch` statement, since we only want one release label to trigger.

```go
switch {
case  c.UpgradeProviderVersion && len(target.GHIssues) > 0:
...
case c.MaintenancePatch && !c.UpgradeProviderVersion:
...
}

File: upgrade/steps_helpers.go [hunk 2]
```diff
@@ -4,17 +4,16 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/Masterminds/semver/v3"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"golang.org/x/mod/modfile"
+	"golang.org/x/mod/module"
```
iwahbe, line 10: Can we undo this change?

upgrade/steps_test.go
```diff
@@ -266,3 +266,93 @@ func TestReleaseLabel(t *testing.T) {
...
+			GoPath: "/Users/myuser/go",
+		}
+
+		repo := ProviderRepo{
+			org:  "pulumi",
+			name: "pulumi-cloudinit",
+		}
+
+		maintenanceRelease(context.Wrap(ctx),
+			repo)
```
iwahbe, line 310:
```suggestion
		maintenanceRelease(context.Wrap(ctx), repo)
```

upgrade/steps_test.go
```diff
@@ -266,3 +266,93 @@ func TestReleaseLabel(t *testing.T) {
...
```
iwahbe, line 355:
```suggestion
		maintenanceRelease(context.Wrap(ctx), repo)
```

upgrade/steps.go
```diff
@@ -261,6 +262,39 @@ func UpgradeProviderVersion(
 	return step.Combined("Update TF Provider", steps...)
 }
 
+var maintenanceRelease = stepv2.Func10E("Check if we should release a maintenance patch", func(
+	ctx context.Context,
+	repo ProviderRepo,
+) error {
```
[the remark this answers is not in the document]
guineveresaenger: I agree and this is what I did at first; however [the rest of this codebase does not](https://github.com/pulumi/upgrade-provider/blob/bb4318b313e32435d964af6292b4a72506fc1e50/upgrade/steps.go#L1016) so I've followed that pattern here for consistency.

The test works fine; at least insofar as testing `ctx.MaintenancePatch`.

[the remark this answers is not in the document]
iwahbe: The rest of the codebase uses `ctx` as an implicit return because it was written against `step` (v1) which did not allow returning values.

We are trying to migrate over to a style of explicit return values.

upgrade/steps_test.go
```diff
@@ -296,3 +296,92 @@ func TestParseUpstreamProviderOrgFromModVersion(t *testing.T) {
...
+		}
+
+		result := maintenanceRelease(context.Wrap(ctx), repo)
+		assert.NotNil(t, result)
+	})
+}
+
+func TestCheckMaintenancePatchExpiredCadence(t *testing.T) {
+
+	simpleReplay(t, jsonMarshal[[]*step.Step](t, `[
```
iwahbe, line 346: This will need a rebase, since this function was renamed.

iwahbe (approved):
Needs a rebase, but otherwise LGTM

PR #230, opened 2024-01-08
Update README.md

dirien:
Add `upstream-provider-org` to Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,20 +53,38 @@ Usage:
   upgrade-provider <provider> [flags]
 
 Flags:
+      --allow-missing-docs              If true, don't error on missing docs during tfgen.
+                                        This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.
       --experimental                    Enable experimental features, such as auto token mapping and auto aliasing
   -h, --help                            help for upgrade-provider
+      --java-version string             The version of pulumi-java-gen to target.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,20 +53,38 @@ Usage:
...
 
 Flags:
+      --allow-missing-docs              If true, don't error on missing docs during tfgen.
+                                        This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.
       --experimental                    Enable experimental features, such as auto token mapping and auto aliasing
   -h, --help                            help for upgrade-provider
+      --java-version string             The version of pulumi-java-gen to target.
       --kind strings                    The kind of upgrade to perform:
-                                        - "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code".
+                                        - "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code,pf".
```
ericrudder, line 62: should we alphabetize list (both in the "shorthand" list AND the entire --kind list?)

dirien: @iwahbe should I change the code in this PR for this or do you prefer an extra PR?

guineveresaenger (approved):
Thank you for this update Engin!

The changes look good to me as-is. We can alphabetize separately.

README.md
iwahbe: We can do it in a follow up PR.

PR #233, opened 2024-01-08
Get upstream org from goMod object

### LINKED ISSUE #231: Improve error message when `upstream-provider-org` is required
## Hello!

## Issue details

Currently, when `upstream-provider-org` is missing (but should be present), the error looks like this:

```
-- failed ---
  - X Lookup Tag SHA: exit status 128:
remote: Repository not found.
fatal: repository 'https://github.com//terraform-provider-port-labs/' not found
```

When we are unable to find the upstream repo, we should fail with a clear and actionable error.

guineveresaenger:
We were reading the upstream org from a place where we didn't set it.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -182,7 +182,7 @@ func UpgradeProviderVersion(
 		// If they are versioning correctly, `go mod tidy` will resolve the SHA to a tag.
 		steps = append(steps,
 			step.F("Lookup Tag SHA", func(context.Context) (string, error) {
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
I can see how this got through code review :(. Let's make sure that the following refactor includes a test that pins the correct behavior here!

iwahbe:
Can we remove the fixes for #231, since it doesn't actually make the error message better (and we should still do that).

guineveresaenger:
oh, sorry - yeah; I thought 231 was a bug issue. Removed!

PR #234, opened 2024-01-08
Fix help formatting

iwahbe:
I noticed this when viewing #230.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"go/build"
 	"os"
+	"strconv"
 	"strings"
 
 	semver "github.com/Masterminds/semver/v3"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -178,22 +179,27 @@ If no version is passed, the pulumi/{pkg,sdk} version will track the bridge`)
 
 	cmd.PersistentFlags().BoolVar(&context.InferVersion, "pulumi-infer-version", false,
 		`Use our GH issues to infer the target upgrade version.
-		If both '--target-version' and '--pulumi-infer-version' are passed,
-		we take '--target-version' to cap the inferred version. [Hidden behind PULUMI_DEV]`)
+If both '--target-version' and '--pulumi-infer-version' are passed,
+we take '--target-version' to cap the inferred version. [Hidden behind PULUMI_DEV]`)
```
guineveresaenger, line 183: This help text in actuality hidden no matter what due to the `MarkHidden` below.

We could probably make this message less confusing - as written, it makes it sound like the _functionality_ is hidden behind PULUMI_DEV, which is not the case.

main.go
```diff
@@ -178,22 +179,27 @@ If no version is passed, the pulumi/{pkg,sdk} version will track the bridge`)
...
+	kindMsg := `The kind of upgrade to perform:
+
+- "all": Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code,pf".
+- "bridge": Upgrade the bridge only.
 - "provider": Upgrade the upstream provider only.
 - "pf": Upgrade the Plugin Framework only.
-- "code":     Perform some number of code migrations.
-- "check-upstream-version": Determine if we need to upgrade the upstream provider. For use in CI only."`)
+- "code": Perform some number of code migrations.`
+	if b, _ := strconv.ParseBool(os.Getenv("PULUMI_DEV")); b {
```
guineveresaenger, line 197: Why do we want to hide this help text? I'm not sure we should.

iwahbe: I thought that we didn't want external users to use this since we are not sure we want to maintain the functionality as is.

main.go
iwahbe: Fixed in https://github.com/pulumi/upgrade-provider/pull/234/commits/cdf3287eac7078a344b0c9cc76338a3c36e8e74b

main.go
guineveresaenger: gotcha - yeah, I was wondering if that was why. I do think the main change maybe would be to move this into its own flag rather than under `--kind` and it's certainly not harmful for third-party users to use.

I don't feel _super_ strongly here since it's only the help text that's hidden and not the functionality that's unset, but in general I find gating things behind an env var unintuitive and difficult to maintain - so, once we have decided on this flag's final form, I'd want to un-hide it.

guineveresaenger (approved):

PR #235, opened 2024-01-08
Refactor upstream org logic

guineveresaenger:
This pull request is a refactor.

When we obtain upstream dependencies, we will need to use the dependency path in the go.mod.

When we obtain upstream releases or tags, we need to use the GitHub URL org/repo path. 

This PR disambiguates the two by:

- removing `upstreamProviderOrg` from the `GoMod` object 
- Setting `upstreamProviderOrg` in the `ctx` if not provided via upgrade config
- Using `ctx.UpstreamProviderOrg` in all appropriate places
- Refactoring some functions to reflect that the upstream org information is available via the `ctx` object and needs no extra parameter.


upgrade/steps_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"golang.org/x/mod/module"
```
iwahbe, line 7: This should be moved out of the std level import block.

upgrade/steps.go
```diff
@@ -1092,3 +1091,13 @@ var fetchLatestJavaGen = stepv2.Func03("Fetching latest Java Gen", func(ctx cont
 	stepv2.SetLabel(ctx, latestJavaGen.String())
 	return currentJavaGen, latestJavaGen.String(), true
 })
+
+var getUpstreamProviderOrg = stepv2.Func11("Get UpstreamOrg from go.mod", func(ctx context.Context, upstreamMod module.Version) string {
```
iwahbe, line 1095: This function isn't getting anything from a `go.mod` file, it's parsing from `module.Version`. 

```suggestion
var parseUpstreamProviderOrg = stepv2.Func11("Parse Upstream Provider Org", func(ctx context.Context, upstreamMod module.Version) string {
```

upgrade/steps_test.go
```diff
@@ -266,3 +267,33 @@ func TestReleaseLabel(t *testing.T) {
...
+        }
+]`), func(ctx context.Context) {
+		context := &Context{
+			GoPath:               "/Users/myuser/go",
+			UpstreamProviderName: "terraform-provider-datadog",
+			UpstreamProviderOrg:  "",
+		}
+		result := getUpstreamProviderOrg(context.Wrap(ctx), upstreamVersion)
+		assert.NotNil(t, result)
+		assert.Equal(t, result, "testing-org")
```
iwahbe, line 297: Since this function is run in the context of a replay, the replay asserts on its inputs and outputs. We don't need the asserts here.

```suggestion
		getUpstreamProviderOrg(context.Wrap(ctx), upstreamVersion)
```

P.S. Since `result` is of type `string`, it *cannot* be `nil`. The `assert.NotNil` cannot fail, regardless of the returned value of `getUpstreamProviderOrg`.

upgrade/upgrade_test.go
```diff
@@ -3,15 +3,15 @@ package upgrade
 import (
 	"context"
 	"encoding/json"
+	"github.com/Masterminds/semver/v3"
+	"golang.org/x/mod/module"
```
iwahbe, line 7: Can we undo this change?

iwahbe (approved):
LGTM

PR #236, opened 2024-01-08
Enable gci linter with automatic fix. Apply lint fixes.

guineveresaenger:
If we want to standardize import grouping and sorting, we should enforce it in code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,10 @@ install:
 	go install github.com/pulumi/upgrade-provider
 
 lint:
-	golangci-lint run -E gci --fix
+	golangci-lint run
+
+lint.fix:
+	golangci-lint run --fix
 
 test:
 	go test -v ./...
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ install:
 	go install github.com/pulumi/upgrade-provider
 
 lint:
-	golangci-lint run
+	golangci-lint run -E gci --fix
 
 test:
 	go test -v ./...
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,13 @@
+linters:
+  enable:
+    - gci
+
+linters-settings:
+  gci:
+    sections:
+      - standard # Standard section: captures all standard packages.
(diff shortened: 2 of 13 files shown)


Makefile
```diff
@@ -11,7 +11,7 @@ install:
 	go install github.com/pulumi/upgrade-provider
 
 lint:
-	golangci-lint run
+	golangci-lint run -E gci --fix
```
iwahbe, line 14: My prior is that `make lint` shouldn't make code changes. Instead changes should be delegated to another command, such as `make lint.fix`.

`-E` -> `--enable` because I didn't know what `-E` meant, but `--enable` is readable.

```suggestion
	golangci-lint run --enable gci
```

iwahbe (commented):
I 100% agree that `gci` should flag these. 

The style we normally use is:

```
import (
    std

    default

    github.com/pulumi/upgrade-provider/*  // (local) imports
)
```

I believe we can configure `gci` to support this.

Makefile
guineveresaenger: Yeah, that's totally fair - I think a good state is where we can run `make lint` and see necessary changes.

guineveresaenger:
Do we have these standards documented somewhere? I looked into it and we can definitely configure custom rules.

iwahbe:
> Do we have these standards documented somewhere? I looked into it and we can definitely configure custom rules.

Not that I'm aware of. I see two patterns in our code bases:
1. The one that I mentioned above. ([example](https://github.com/pulumi/pulumi/blob/644fa55ef5dee5b45434a1c1a350c1f533dd9bcf/pkg/engine/plugin_host.go#L17-L27))
2. A variation, where there is an additional block for `github.com/pulumi/*`. ([example](https://github.com/pulumi/pulumi-terraform-bridge/blob/71d95766741ca6d65b88fed313669c658ab45fd7/pkg/tfbridge/auto_aliasing.go#L17-L29))

I'm fine with either.

iwahbe (approved):
Awesome! LGTM!

PR #237, opened 2024-01-11
allow only upgrading pulumi

dixler:
This adds a kind to only upgrade the pulumi version.

This is used by pulumi/pulumi to run downstream codegen tests and the PRs should not be merged.

Part of https://github.com/pulumi/pulumi/issues/15140

### DIFF
--- a/main.go
+++ b/main.go
@@ -129,6 +129,7 @@ func cmd() *cobra.Command {
 					set(&context.UpgradeCodeMigration)
 				case "pf":
 					set(&context.UpgradePfVersion)
+				case "pulumi":
 				case "check-upstream-version":
 					if targetVersion != "" {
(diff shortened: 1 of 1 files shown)


VenelinMartinov (approved):

main.go
```diff
@@ -129,6 +129,7 @@ func cmd() *cobra.Command {
 					set(&context.UpgradeCodeMigration)
 				case "pf":
 					set(&context.UpgradePfVersion)
+				case "pulumi":
```
iwahbe, line 132: It would be really nice to have a comment here that explains what this does. Right now, it looks like `"pulumi"` does nothing.

I'm assuming that this is effetely a way to be able to avoid the `"all"` kind without passing any other kind.

PR #239, opened 2024-01-16
Fail fast on --kind empty

t0yv0:
Before the change `upgrade provider --kind=""` would proceed with "No changes required" success message. I believe it is better to fail fast.


VenelinMartinov:
I'm confused why it does not error here: https://github.com/pulumi/upgrade-provider/blob/d846fbbe9314cf08280f499bbc3fd98291e34e52/main.go#L115

VenelinMartinov (approved):

blampe (approved):

iwahbe (approved):

PR #240, opened 2024-01-30
Formatting

VenelinMartinov:
Will reduce noise in my next PR.

I don't have an explanation why my gofmt yields diffs to CI gofmt.


iwahbe (approved):
Is it possible that you are using `gofumpt` instead of `gofmt`? That would explain the stricter formatting.

PR #241, opened 2024-01-30
Allow upgrading only tf sdk and fix TF SDK upgrades

VenelinMartinov:
This fixes TF SDK upgrades and allows upgrade-provider to make PRs to only upgrade that.

Example PR: https://github.com/pulumi/pulumi-azuredevops/pull/281

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -137,7 +137,7 @@ var ensureUpstreamRepo = stepv2.Func11("Ensure Upstream Repo", func(ctx context.
 	if !repoExists {
 		stepv2.Func10("Downloading", func(ctx context.Context, path string) {
 			targetDir := stepv2.NamedValue(ctx, "Target Dir", filepath.Dir(path))
-			stepv2.MkDirAll(ctx, targetDir, 0700)
+			stepv2.MkDirAll(ctx, targetDir, 0o700)
 			stepv2.Cmd(ctx, "git", "clone", fmt.Sprintf("https://%s.git", repoPath), path)
 		})(ctx, expectedLocation)
 	}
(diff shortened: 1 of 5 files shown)


upgrade/steps.go
```diff
@@ -1021,7 +1025,9 @@ var planPluginSDKUpgrade = stepv2.Func12E("Planning Plugin SDK Upgrade", func(
 	currentBranch, ok := refs.labelOf(currentRef)
 	if !ok {
 		// use latest versioned branch
```
VenelinMartinov, line 1027: Looks like this comment was inaccurate - it would previously return "" which got picked up on https://github.com/pulumi/upgrade-provider/blob/1b9fe00eda2b8a52416099d3903a850a689d4699/upgrade/upgrade_provider.go#L213 to mean don't upgrade.

upgrade/steps.go
```diff
@@ -535,6 +537,8 @@ var getWorkingBranch = stepv2.Func41E("Working Branch Name", func(ctx context.Co
 		return ret("upgrade-pulumi-version-to-%s", c.TargetPulumiVersion)
 	case c.UpgradeJavaVersion:
 		return ret("upgrade-java-version-to-%s", c.JavaVersion)
+	case c.UpgradeTFSDKVersion:
+		return ret("upgrade-tf-sdk-version-to-%s", c.TargetTFSDKRef)
```
iwahbe, line 541: As I understand it, we will only hit this case when performing a bridge upgrade where the bridge is already up to date but a new version of the TF SDK has been released. Is that correct?

I'm not sure we actually want to upgrade the TF SDK in that case. It seems like we only need to trigger a TF SDK upgrade when we are upgrading the bridge.

VenelinMartinov:
superseded by https://github.com/pulumi/upgrade-provider/pull/242

Issue #243, opened 2024-01-31
Do not upgrade forked terraform-plugin-sdk ahead of the bridge


t0yv0:
### What happened?

Per conversation in https://github.com/pulumi/upgrade-provider/pull/242/files it appears that maybe/possibly we're upgrading the fork of terraform-plugin-sdk to the latest from GitHub. That's may be a good option to have but it's not a good default. By default we should be trying to match whatever the bridge has when upgrading the bridge, because that is the version the bridge was tested with. Unfortunately go tooling does not make this trivial, but something along the lines of finding bridge sources and finding the go mod replace directive from those sources that the bridge was built with could do the trick. If helpful we can lift this to a text file in the bridge somehow.

### Example

N/A

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
https://github.com/pulumi/pulumi-gcp/pull/1638/files#diff-0301265a2d4321ca487bbd7ba523f69633171cef680c0114e7ca8d2c34176b73R19 another one.. I was updating GCP to do a simple minor update and looks like the terraform-plugin-sdk is updating as part of it. That's a bit suspect.

t0yv0:
OK this is causing failure to pre-test bridge upgrade with a given sdk version

--target-bridge-version: 45cdb2507f387f34af187ca4021076549cbbf78e

https://github.com/pulumi/pulumi-github/issues/559 and more

PR #248, opened 2024-02-08
Ensure sdkv2 replace mirrors the bridge replace

t0yv0:
Fixes https://github.com/pulumi/upgrade-provider/issues/243

After this change the terraform SDK v2 override follows the replace from the selected bridge version exactly. I think this is what we want - this is compensating for the inability of Go tooling to propagate the replace to dependencies.

The change also starts to unconditionally insert this replace. This is because on the latest bridge tfgen somehow depends on sdkv2 and even providers such as pulumi-random that presumably do not need sdkv2 now depend on it and need the replace accordingly, see https://github.com/pulumi/pulumi-random/issues/687 for example.

### DIFF
--- a/upgrade/http.go
+++ b/upgrade/http.go
@@ -1,0 +1,73 @@
+// HTTP-related effects.
+package upgrade
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+)
+
+// Use this function in steps to perform HTTP GET.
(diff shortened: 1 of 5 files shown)


guineveresaenger (approved):
This looks fine; but do we need the retry logic? I assume this is to make it more robust in CI.

t0yv0:
Yeah. Well these days pretty much any time you do http logic you need to retry.

PR #250, opened 2024-02-09
Fix planning SDK update when bridge does not change

t0yv0:
Quick fix for a panic introduced in https://github.com/pulumi/upgrade-provider/issues/243

Would like to merge to prevent p1s, follow up with tests later.

Verified this works on pulumi-azuread.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -963,6 +963,11 @@ var planProviderUpgrade = stepv2.Func41E("Plan Provider Upgrade", func(ctx conte
 	return upgradeTarget, nil
 })
 
+// If a bridge update is needed, will return the concrete Ref (never returns "latest").
+//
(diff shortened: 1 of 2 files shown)


upgrade/steps.go
```diff
@@ -963,6 +963,11 @@ var planProviderUpgrade = stepv2.Func41E("Plan Provider Upgrade", func(ctx conte
 	return upgradeTarget, nil
 })
 
+// If a bridge update is needed, will return the concrete Ref (never returns "latest").
+//
+// If the bridge is up to date, will return nil and unset the following flag:
+//
+//	GetContext(ctx).UpgradeBridgeVersion = false
```
guineveresaenger, line 970: thank you!

guineveresaenger (approved):

PR #252, opened 2024-02-29
Allow title prefix arg

VenelinMartinov:
This will allow us to prefix downstream test PRs with [TEST] or something so that it's more obvious which PRs are tests and which are real upgrades.


iwahbe (approved):
Looks good.

PR #261, opened 2024-05-09
Add needs release arg

VenelinMartinov:
This can be used to trigger releases of all providers on a bridge release.

### DIFF
--- a/main.go
+++ b/main.go
@@ -256,6 +256,8 @@ This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.`)
 	cmd.PersistentFlags().StringVar(&context.PRTitlePrefix, "pr-title-prefix", "",
 		`The prefix to insert in the generated pull request title.`)
 
+	cmd.PersistentFlags().BoolVar(&context.NeedsReleasePatch, "needs-release-patch", false, "setting this field to True will trigger a patch release after the PR is merged.")
+
 	return cmd
 }
 
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -345,7 +345,11 @@ var InformGitHub = stepv2.Func70E("Inform Github", func(
 			}
 		// On non-upstream upgrades, we will create a patch release label
 		// if the provider hasn't been released in 8 weeks.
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -256,6 +256,8 @@ This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.`)
 	cmd.PersistentFlags().StringVar(&context.PRTitlePrefix, "pr-title-prefix", "",
 		`The prefix to insert in the generated pull request title.`)
 
+	cmd.PersistentFlags().BoolVar(&context.NeedsReleasePatch, "needs-release-patch", false, "setting this field to True will trigger a patch release after the PR is merged.")
```
iwahbe, line 259: Instead of `--needs-release-patch=true`, can we have an enum flag `--needs-release=patch`.

upgrade/steps.go
```diff
@@ -345,7 +345,11 @@ var InformGitHub = stepv2.Func70E("Inform Github", func(
 			}
 		// On non-upstream upgrades, we will create a patch release label
 		// if the provider hasn't been released in 8 weeks.
-		case c.MaintenancePatch && !c.UpgradeProviderVersion:
+		case (c.MaintenancePatch && !c.UpgradeProviderVersion):
```
iwahbe, line 348: We don't need this change.
```suggestion
		case c.MaintenancePatch && !c.UpgradeProviderVersion:
```

upgrade/steps.go
```diff
@@ -345,7 +345,11 @@ var InformGitHub = stepv2.Func70E("Inform Github", func(
...
+			addLabels = []string{"--label", "needs-release/patch"}
+
+		// If we have the needs release flag, we make a patch release.
+		case c.NeedsReleasePatch:
```
iwahbe, line 352: We should put `c.NeedsReleasePatch` before the normal release switch statement. Otherwise we could get a scenario where the `case c.UpgradeProviderVersion && len(target.GHIssues) > 0:` branch triggers but no label is added, even though `c.NeedsReleasePatch` is true.

iwahbe (commented):
This at least a basic test.

PR #265, opened 2024-05-29
Improve error message around expected go module path format

### LINKED ISSUE #246: Tf plugin sdk upgrades should update replaces in examples folder as well as provider
## Hello!

## Issue details

Some providers have the plugin sdk replace in the examples go.mod, so we should update that too to avoid manual interventions. Examples:

https://github.com/pulumi/pulumi-hcloud/pull/399
https://github.com/pulumi/pulumi-azuredevops/actions/runs/7741246755/job/21111395025?pr=284

### LINKED ISSUE #264: index out of range [1] with length 1 when upgrading meraki provider
### What happened?

```
upgrade-provider pulumi/pulumi-meraki
```

fails with

```
Get Repo Kind
    - /Users/mikhailshilkov/go/src/github.com/pulumi/pulumi-meraki/provider/go.mod: 13050 bytes read
\ [Get Repo Kind]panic: runtime error: index out of range [1] with length 1

goroutine 82 [running]:
github.com/pulumi/upgrade-provider/upgrade.glob..func19({0x102731000?, 0x140000b0150?}, {{0x14000320280?, 0x10228ebc4?}, {0x140003223a8?, 0x1027c45c0?}})
	/Users/mikhailshilkov/go/src/github.com/pulumi/upgrade-provider/upgrade/steps.go:1131 +0x88
github.com/pulumi/upgrade-provider/upgrade.init.Func11[...].func11({{0x14000320280?, 0x4?}, {0x140003223a8?, 0x102280798?}})
	/Users/mikhailshilkov/go/src/github.com/pulumi/upgrade-provider/step/v2/calls.go:160 +0x34
reflect.Value.call({0x102791940?, 0x140001202e8?, 0x1499d1e88?}, {0x10264416a, 0x4}, {0x14000426150, 0x2, 0x0?})
	/opt/homebrew/Cellar/go/1.21.6/libexec/src/reflect/value.go:596 +0x994
reflect.Value.Call({0x102791940?, 0x140001202e8?, 0x140000c2d38?}, {0x14000426150?, 0x0?, 0x200000003?})
	/opt/homebrew/Cellar/go/1.21.6/libexec/src/reflect/value.go:380 +0x94
github.com/pulumi/upgrade-provider/step/v2.run.func2()
	/Users/mikhailshilkov/go/src/github.com/pulumi/upgrade-provider/step/v2/step.go:198 +0x774
created by github.com/pulumi/upgrade-provider/step/v2.run in goroutine 37
	/Users/mikhailshilkov/go/src/github.com/pulumi/upgrade-provider/step/v2/step.go:158 +0x1f8
```

This is because the upstream provider has a unexpected module name of `terraform-provider-meraki` without the `github.com/<org>` part. Our parsing does not expect that and fails.

We should probably stop panic'ing and give a user-friendly error.

Relatedly, how do I fix this for this specific provider? Specify `upstream-provider-org: cisco-open` somewhere in ci-mgmt?

### Example

see above

### Output of `pulumi about`

irrelevant

### Additional context

_No response_

### Contributing

guineveresaenger:
Fixes #264.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -1129,10 +1129,10 @@ var parseUpstreamProviderOrg = stepv2.Func11E("Get UpstreamOrg from module versi
 	// The second chunk is the org name.
 
 	// Verify that the token is of valid format
-	goModRegexp := regexp.MustCompile("github.com/[a-zA-Z0-9-]*/[a-zA-Z0-9-]*")
+	goModRegexp := regexp.MustCompile("[a-zA-Z0-9-.]*/[a-zA-Z0-9-_.]*/[a-zA-Z0-9-]*")
 
 	if !goModRegexp.MatchString(upstreamMod.Path) {
-		return "", fmt.Errorf(fmt.Sprintf("Invalid upstream module format: expected format github.com/${org}/${repo}/${path} but got %s", upstreamMod.Path))
+		return "", fmt.Errorf("invalid upstream module format: expected format github.com/${org}/${repo}/${path} but got %s", upstreamMod.Path)
 	}
 	tok := strings.Split(modPathWithoutVersion(upstreamMod.Path), "/")
 	return tok[1], nil
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -1121,12 +1121,19 @@ var fetchLatestJavaGen = stepv2.Func03("Fetching latest Java Gen", func(ctx cont
 	return currentJavaGen, latestJavaGen.String(), true
 })
 
-var parseUpstreamProviderOrg = stepv2.Func11("Get UpstreamOrg from module version", func(ctx context.Context, upstreamMod module.Version) string {
+var parseUpstreamProviderOrg = stepv2.Func11E("Get UpstreamOrg from module version", func(ctx context.Context, upstreamMod module.Version) (string, error) {
 	// We expect tokens to be of the form:
 	//
 	//	github.com/${org}/${repo}/${path}
 	//
 	// The second chunk is the org name.
+
+	// Verify that the token is of valid format
+	goModRegexp := regexp.MustCompile("github.com/[a-zA-Z0-9-]*/[a-zA-Z0-9-]*")
+
+	if !goModRegexp.MatchString(upstreamMod.Path) {
+		return "", fmt.Errorf(fmt.Sprintf("Invalid upstream module format: expected format github.com/${org}/${repo}/${path} but got %s", upstreamMod.Path))
+	}
 	tok := strings.Split(modPathWithoutVersion(upstreamMod.Path), "/")
-	return tok[1]
+	return tok[1], nil
 })
--- a/upgrade/steps_test.go
+++ b/upgrade/steps_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/mod/module"
 
 	"github.com/pulumi/upgrade-provider/step/v2"
 )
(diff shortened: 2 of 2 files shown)


upgrade/steps.go
```diff
@@ -1121,12 +1121,19 @@ var fetchLatestJavaGen = stepv2.Func03("Fetching latest Java Gen", func(ctx cont
...
-var parseUpstreamProviderOrg = stepv2.Func11("Get UpstreamOrg from module version", func(ctx context.Context, upstreamMod module.Version) string {
+var parseUpstreamProviderOrg = stepv2.Func11E("Get UpstreamOrg from module version", func(ctx context.Context, upstreamMod module.Version) (string, error) {
 	// We expect tokens to be of the form:
 	//
 	//	github.com/${org}/${repo}/${path}
 	//
 	// The second chunk is the org name.
+
+	// Verify that the token is of valid format
+	goModRegexp := regexp.MustCompile("github.com/[a-zA-Z0-9-]*/[a-zA-Z0-9-]*")
```
iwahbe, line 1132: I know your following the helpful comment, but I'm pretty sure that the comment is wrong. 
```suggestion
	goModRegexp := regexp.MustCompile("[\.a-zA-Z0-9-]*/[a-zA-Z0-9-]*/[a-zA-Z0-9-]*")
```

Specifically, I think this will break `pulumi-gitlab`:

https://github.com/pulumi/pulumi-gitlab/blob/5e8f66b41c8250bcd117a1a45333db3b098bcddd/provider/go.mod#L11

upgrade/steps.go
```diff
@@ -1121,12 +1121,19 @@ var fetchLatestJavaGen = stepv2.Func03("Fetching latest Java Gen", func(ctx cont
...
+
+	if !goModRegexp.MatchString(upstreamMod.Path) {
+		return "", fmt.Errorf(fmt.Sprintf("Invalid upstream module format: expected format github.com/${org}/${repo}/${path} but got %s", upstreamMod.Path))
```
iwahbe, line 1135: `fmt.Errorf` already does the formatting part (notice the trailing `f`), so we don't need to call `fmt.Sprintf` as well.
```suggestion
		return "", fmt.Errorf("Invalid upstream module format: expected format github.com/${org}/${repo}/${path} but got %s", upstreamMod.Path)
```

upgrade/steps.go
rquitales, line 1132: Q: do we need to make this less restrictive to allow underscores and other characters?

upgrade/steps.go
rquitales, line 1135: Regex only asserts `github.com/${org}/${repo}`, do we need to also regex check the path component?

blampe:
How does this address #246? Don't we still have a replace problem?

upgrade/steps.go
guineveresaenger: The `*` does include any trailing path component and I'm not sure it's useful to make it any stricter. I verified that this code gets hit for [the GitHub provider's upstream module](https://github.com/integrations/terraform-provider-github/blob/main/go.mod#L1).

upgrade/steps.go
guineveresaenger: Great question! I'm not aware of any repos that have those at this time. But the point here is to check for a complete module path so I don't see any reason not to add the.

For reference:

[Underscores are not allowed in TF module names](https://developer.hashicorp.com/terraform/language/style#module-repository-names)

Underscores are not allowed in GH org names (I couldn't find docs, but trying to set up an organization proved this), but [GitLab seems to allow it](https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-usernames-project-and-group-names), and allows dots as well.
Dots seem to be allowed in GitHub also, although GitHub substitutes those for dashes.

guineveresaenger: Indeed! thank you for catching that.

iwahbe (approved):
I'd love to see a simple unit test here to validate `goModRegexp`. 

Something like this:

https://github.com/pulumi/upgrade-provider/blob/9692355a2b2548ed824eabe188974845d4f2c789/upgrade/steps_test.go#L219

It doesn't need to be as elaborate as ☝️.

PR #271, opened 2024-07-02
Adjust rebase procedure

danielrbradley:
- Based on new targets from https://github.com/pulumi/ci-mgmt/pull/1020
- Perform a real rebase, then extract patches & the new base.

Needs to wait until https://github.com/pulumi/ci-mgmt/pull/1020 has been rolled out to all providers.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -164,25 +164,12 @@ func UpgradeProviderVersion(
 			step.Cmd("git", "fetch", "--tags").In(&upstreamDir),
 			// We need to remove any patches to so we can cleanly pull the next upstream version.
 			step.Cmd("git", "reset", "HEAD", "--hard").In(&upstreamDir),
-			// Load patches into a 'pulumi-patch' branch.
-			// This also creates a "checkout-base" branch which marks where to extract patches back to.
-			step.Cmd("make", "upstream.checkout").In(&repo.root),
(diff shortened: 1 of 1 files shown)


step/step.go
```diff
@@ -84,6 +84,8 @@ func Cmd(name string, args ...string) Step {
 		output = string(out)
 		if exit, ok := err.(*exec.ExitError); ok {
 			err = fmt.Errorf("%s:\n%s", err.Error(), string(exit.Stderr))
+		} else if err != nil {
+			err = fmt.Errorf("%s", err.Error())
```
thomas11, line 88: What does this achieve? `.Error()` should be printed by default, no?

thomas11 (approved):

step/step.go
danielrbradley: Not 100% sure ... just copying from above to handle the case where the error is not an exit error.

Would be great to figure out how to test this kind of machinery better.

thomas11: I think the case above just tries to concatenate the error with stderr. Without stderr, we should be able to just return `err`.

danielrbradley: Oh yeah you're right 

I'm finding this code hard to follow as it seems like it always actually returns an empty string ... but the actual useful return is done by setting the `output` variable ... which is set in the `.Return(&output)`. Guess this isn't the source of the lost console output then 😞

PR #275, opened 2024-08-19
Clear examples cache during major upgrades

guineveresaenger:
Update: Apologies for the delay here; see https://github.com/pulumi/pulumi-terraform-bridge/issues/2323.

https://github.com/pulumi/pulumi-confluentcloud/pull/559 failed because the upgrade process was using an old examples cache locally.
As a result, the Go SDK was generated with old version paths in the examples code, which failed the Clean Worktree check.
The upgrade-provider tool should remove the cache to ensure we're generating examples correctly.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -673,6 +673,9 @@ var majorVersionBump = stepv2.Func30("Increment Major Version", func(
 	setEnv(ctx, "VERSION_PREFIX", nextMajorVersion)
 
 	addVersionPrefixToGHWorkflows(ctx, repo, nextMajorVersion)
+
+	// Remove examples cache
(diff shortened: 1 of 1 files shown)


t0yv0:
Suggest fixing in the bridge instead since some interaction happens via `make tfgen` instead of via upgrade-provider, that'd be more definitive.

danielrbradley (approved):
Seems reasonable for a short-term fix, if we can't land a fix to the cache key in the next couple of weeks

iwahbe (commented):
I'd much rather prioritize adding the major version to the cache in the bridge then fix it here. Unless this problem comes up regularly, I would suggest just fixing the bridge.

If you do need to merge, please add a TODO linking the step to the bridge bug.

t0yv0:
I'm with Daniel, ship the workaround. We'll fix the root cause later.

PR #278, opened 2024-10-01
Upgrade to Go 1.23 and golang.org/x to 0.21.0

flostadler:
This is necessary for upgrade-provider to detect new godebug directives like `tlskyber`.
That godebug directive is used in pulumi-aws to work around an issue with AWS firewalls. Right now upgrade-provider fails when parsing the `go.mod` file because it doesn't know about the `tlskyber` directive.

I verified this correctly works by executing an upstream check on the aws repo:
```
GH_TOKEN="$(gh auth token)" ./../upgrade-provider/bin/upgrade-provider "pulumi/pulumi-aws" --repo-path . --upstream-provider-name terraform-provider-aws --kind=check-upstream-version
```

Relates to https://github.com/pulumi/pulumi-aws/issues/4586


iwahbe (approved):

iwahbe:
I think you need to bump the linter version as well.

upgrade/http.go
```diff
@@ -29,7 +29,9 @@ type httpHandler interface {
 	getHTTP(url string) ([]byte, error)
 }
 
-var httpHandlerKey = struct{}{}
+type httpContextKey struct{}
```
flostadler, line 32: Addresses linter warning SA1029: should not use empty anonymous struct as key for value; define your own type to avoid collisions

upgrade/migrations.go
```diff
@@ -24,7 +24,7 @@ const (
 func AutoAliasingMigration(resourcesFilePath, providerName string) (bool, error) {
 	// Create the AST by parsing src
 	fset := token.NewFileSet()
-	file, err := parser.ParseFile(fset, resourcesFilePath, nil, parser.ParseComments)
+	file, err := parser.ParseFile(fset, resourcesFilePath, nil, parser.ParseComments|parser.SkipObjectResolution)
```
flostadler, line 27: Syntactic object resolution is deprecated now including `ast.Object`, see: https://pkg.go.dev/go/ast#Object.

We don't need object resolution because we're doing ast manipulation and not analysis or resolving symbols across scopes.

upgrade/migrations.go
```diff
@@ -96,7 +96,7 @@ func AutoAliasingMigration(resourcesFilePath, providerName string) (bool, error)
 					}
 					c.InsertBefore(&ast.AssignStmt{
 						Tok: tok,
-						Lhs: []ast.Expr{&ast.Ident{Name: "err", Obj: &ast.Object{Kind: ast.Var, Name: "err"}}},
+						Lhs: []ast.Expr{&ast.Ident{Name: "err"}},
```
flostadler, line 99: `Ident.Obj` is deprecated now. It wasn't needed for our use cases here.

upgrade/upgrade_provider.go
```diff
@@ -202,7 +202,7 @@ func UpgradeProvider(ctx context.Context, repoOrg, repoName string) (err error)
 			return err
 		}
 		defer func() {
-			fmt.Printf("\n\n" + colorize.Warn("Major Version Updates are not fully automated!") + "\n")
+			fmt.Printf("\n\n%s\n", colorize.Warn("Major Version Updates are not fully automated!"))
```
flostadler, line 205: Fixed: non-constant format string in call to fmt.Printf

upgrade/kind.go
```diff
@@ -98,7 +98,7 @@ var getRepoKind = stepv2.Func11E("Get Repo Kind", func(ctx context.Context, repo
 	if err != nil {
 		return nil, fmt.Errorf("%s: %w", bridgeMissingMsg, err)
 	} else if !ok {
-		return nil, fmt.Errorf(bridgeMissingMsg)
+		return nil, fmt.Errorf("%s", bridgeMissingMsg)
```
flostadler, line 101: Fixed: non-constant format string in call to fmt.Errorf

iwahbe (approved):
Thanks for fixing the lints.

PR #282, opened 2024-10-25
Allow upgrade-provider to be run outside of CI

### LINKED ISSUE #280: Make issue author configurable
This should work for third party providers which don't have pulumibot.

danielrbradley:
Filtering via @me assumes that the user is always the same when creating issues. This might not be true if a user runs this tool locally. This also breaks when the author is an app such as github-actions as this must be specified via the `--app` option instead of `--author`.

As an alternative approach, GitHub will index for search any word within the issue body, even hidden within an HTML comment. We'll include the word "pulumiupgradeproviderissue" in the HTML comment to allow listing all upgrade issues easily via search. Once we've listed candidate issues, we'll check client-side for an exact matching title.

Here's an example issue where I've manually added this hidden comment to test searching:
- https://github.com/pulumi/pulumi-aws/issues/4652

Issue description preview

We'll leave the existing "@me" based search in as a fallback until we can be confident that these tokens will be present in relevant issues.

Resolves #280

### DIFF
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -722,11 +722,14 @@ var createUpstreamUpgradeIssue = stepv2.Func30E("Ensure Upstream Issue", func(ct
 func upgradeIssueExits(ctx context.Context, title, repoOrg, repoName string) (bool, error) {
 	var found bool
 	var err error
-	// Search through existing "pulumiupgradeproviderissue" issues to see if we've already created one for this version.
+	// TODO: Remove this after we've migrated all issues to the new format.
+	// https://github.com/pulumi/upgrade-provider/issues/284
+	// Fall back to searching through the issues from the current user.
 	found, err = issueExistsForVersion(ctx, title,
 		fmt.Sprintf("--repo=%q", repoOrg+"/"+repoName),
-		fmt.Sprintf("--search=%q", upgradeIssueBodyTemplate),
-		"--state=open")
+		fmt.Sprintf("--search=%q", title),
+		"--state=open",
+		"--author=@me")
 	if err != nil {
 		return false, err
 	}
@@ -734,12 +737,11 @@ func upgradeIssueExits(ctx context.Context, title, repoOrg, repoName string) (bo
 		return true, nil
 	}
 
-	// Fall back to searching through the issues from the current user.
+	// Search through existing pulumiupgradeproviderissue issues to see if we've already created one for this version.
 	found, err = issueExistsForVersion(ctx, title,
 		fmt.Sprintf("--repo=%q", repoOrg+"/"+repoName),
-		fmt.Sprintf("--search=%q", title),
-		"--state=open",
-		"--author=@me")
+		fmt.Sprintf("--search=%q", upgradeIssueBodyTemplate),
+		"--state=open")
 	if err != nil {
 		return false, err
 	}
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -2,6 +2,7 @@ package upgrade
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


iwahbe (changes requested):
I'd like to avoid situations where external users can effect our tooling. Instead of messing around with base64 encoded magic strings (which can be arbitrarily added to issues), can we simply make the `--author` customizable?

Pulumi could specify `--author` as pulumi-bot, others could set app/github-bot, or whatever is appropriate for their org.

That seems like a much simpler solution and I'm struggling to think of the down side.

upgrade/steps_helpers.go
```diff
@@ -671,46 +671,103 @@ var getExpectedTargetFromIssues = stepv2.Func11E("From Issues", func(ctx context
...
+	}
+
+	// Write issue_created=true to GITHUB_OUTPUT, if it exists for CI control flow.
+	if GITHUB_OUTPUT, found := os.LookupEnv("GITHUB_OUTPUT"); found {
+		f, err := os.OpenFile(GITHUB_OUTPUT, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		if _, err := f.WriteString(fmt.Sprintf("issue_created=%t\n", issueAlreadyExists)); err != nil {
```
iwahbe, line 695: After discussion, this needs to be the latest version we want to upgrade to (if the provider is not already on the latest version), regardless of issue creation.

upgrade/steps_helpers.go
```diff
@@ -671,46 +671,103 @@ var getExpectedTargetFromIssues = stepv2.Func11E("From Issues", func(ctx context
...
+		}
+	}
+
+	// We've found an appropriate existing issue, so we'll skip creating a new one.
+	if issueAlreadyExists {
+		return nil
+	}
+
+	// Hide searchable token in the issue body via an HTML comment to help us find this issue later without requiring labels to be set up.
+	hiddenBody := fmt.Sprintf("<!-- %s -->", pulumiupgradeproviderissue)
```
iwahbe, line 706: Lets add a comment (visible to people who look at the opened issue description) that explains why this comment exists.

danielrbradley:
Notes from call:
- We might want to retry the upgrade nightly, so should always output the latest version to try to upgrade to.
- Add more explanation of the tooling & magic labels used for issues.

Rollout plan:
1. Add new magic string and github output.
2. Update CI mgmt to combine jobs via new output.
3. Wait 1 week.
4. Switch issue search to use magic string.

upgrade/steps_helpers.go
```diff
@@ -671,46 +671,103 @@ var getExpectedTargetFromIssues = stepv2.Func11E("From Issues", func(ctx context
...
-	var issues []struct {
-		Title  string `json:"title"`
-		Number int    `json:"number"`
+	return nil
+})
+
+func upgradeIssueExits(ctx context.Context, title, repoOrg, repoName string) (bool, error) {
+	var found bool
+	var err error
+	// Search through existing "pulumiupgradeproviderissue" issues to see if we've already created one for this version.
```
iwahbe, line 722:
```suggestion
	// Search through existing pulumiupgradeproviderissue issues to see if we've already created one for this version.
```

upgrade/steps_helpers.go
```diff
@@ -671,46 +671,103 @@ var getExpectedTargetFromIssues = stepv2.Func11E("From Issues", func(ctx context
...
+		"--state=open")
+	if err != nil {
+		return false, err
 	}
-	err := json.Unmarshal([]byte(searchIssues), &issues)
+	if found {
+		return true, nil
+	}
+
+	// Fall back to searching through the issues from the current user.
```
iwahbe, line 734: Let's add a TODO and an issue for removing this logic.

iwahbe (approved):

ringods (approved):

PR #286, opened 2024-10-31
Remove searching issues by author

### LINKED ISSUE #284: Remove searching issues by author
Once all open issues have the new format which includes `pulumiupgradeproviderissue`, we should remove the historical search for issues based on the author.

See the added TODO statement added in https://github.com/pulumi/upgrade-provider/pull/282

danielrbradley:
Fixes #284
Follow-up to https://github.com/pulumi/upgrade-provider/pull/282

All issues are now created with the searchable term which allows us to avoid using specific users and can instead use the built-in github-actions app principals.

Note: we're insuring there's no old issues open before shipping this so all new issues definitely have the searchable term.

### DIFF
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -720,41 +720,11 @@ var createUpstreamUpgradeIssue = stepv2.Func30E("Ensure Upstream Issue", func(ct
 })
 
 func upgradeIssueExits(ctx context.Context, title, repoOrg, repoName string) (bool, error) {
-	var found bool
-	var err error
-	// TODO: Remove this after we've migrated all issues to the new format.
-	// https://github.com/pulumi/upgrade-provider/issues/284
-	// Fall back to searching through the issues from the current user.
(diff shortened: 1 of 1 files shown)


danielrbradley:
A [quick search shows](https://github.com/search?q=org%3Apulumi+Upgrade+terraform-provider++%22Release+details%22&type=issues&state=open) that all open issues (1 right now) have the new format applied.

flostadler (approved):
LGTM
What's your plan for insuring there's no old issues open before shipping this?

Edit: Should've read your comment... Saw the search now, all good!

danielrbradley:
> LGTM What's your plan for insuring there's no old issues open before shipping this?

@flostadler the new version of the upgrade-provider is already released and now being used by all providers so any new issues will have the new format. [This search](https://github.com/search?q=org%3Apulumi+Upgrade+terraform-provider++%22Release+details%22&type=issues&state=open) shows that there's also no older issues still around with the old format.

iwahbe (approved):
That was fast. LGTM

PR #288, opened 2024-11-04
Simplify identifying provider upgrade issues

danielrbradley:
1. Use search token for finding issues to close
2. Rename variable  - name variable after the type.
3. Remove checking author when finding issues 
   - Now we're using the token for search it's a lot more reliable so we can remove the additional filtering.

Stacked on #287


VenelinMartinov:
@danielrbradley looks like the test workflows have not been updated

VenelinMartinov (approved):

danielrbradley:
Tested locally to create this PR:
- https://github.com/pulumi/pulumi-azure/pull/2596

PR #289, opened 2024-11-05
Make assignee optional

danielrbradley:
PRs and issues can't be assigned to apps, so the default of "@me" will break when being run by the built-in GitHub Actions app from CI when we migrate away from using the single central pulumi-bot account. Therefore, if the assignee is not set explicitly, we will just not set an assignee at all. The assignee should not be critical to any part of the automation.

See also https://github.com/pulumi/ci-mgmt/pull/1127

### DIFF
--- a/main.go
+++ b/main.go
@@ -236,7 +236,7 @@ plugins are used if existing versions are present in the cache.`)
 	cmd.PersistentFlags().StringVar(&context.PrReviewers, "pr-reviewers", "",
 		`A comma separated list of reviewers to assign the upgrade PR to.`)
 
-	cmd.PersistentFlags().StringVar(&context.PrAssign, "pr-assign", "@me",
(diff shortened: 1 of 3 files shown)


flostadler (approved):
Looks good to me! Can't think of a reason that requires pulumi-bot to be the PR assignee

thomas11 (commented):
I'm a bit concerned that PRs without assignee might fall through the cracks. Can we at least assign a team ("providers")?

danielrbradley:
> I'm a bit concerned that PRs without assignee might fall through the cracks. Can we at least assign a team ("providers")?

I think this should be handled via the pr-reviewers rather than the assignee. Currently, assigning to pulumi-bot doesn't help with visibility as no one recieves the assignment notification.

Issue #291, opened 2024-11-06
upgrade-provider no longer adds Fixes keyword and issue to automatic PRs


guineveresaenger:
### What happened?

See example below - it seems that automation-created issues are missing the "Fixes" keyword.

Locally created PRs as https://github.com/pulumi/pulumi-dbtcloud/pull/168 do not have this problem.

### Example

https://github.com/pulumi/pulumi-auth0/pull/687

should have closed 

https://github.com/pulumi/pulumi-auth0/issues/686

but was missing the "Fixes <issue #>" text .

### Output of `pulumi about`

n/a

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

danielrbradley:
Ok, making some headway with this today … I’ve created tags in the upstream xyz provider so we can run upgrades on pulumi-xyz without it failing.

I’ve also added the ability to set the version of upgrade-provider when running in CI for better predictability and to enable testing.

Observations:

1. When using CI to create both the issue and PR, I’m still seeing no “fixes” appear.
2. When I re-run the PR creation locally it adds the “fixes”.
3. When closing existing PRs and re-running in CI, I also see the “fixes”.

Therefore, I think we’re running into an issue with a delay in the GitHub issue searching. I think there’s a delay between opening an issue and being able to search on the contents of the body.

I would guess searching on author doesn’t have this issue because it’s using a simpler indexing strategy which is updated synchronously at the point of saving, but the text body searching is only “eventually consistent”.

Therefore, I could see a few approaches to solving this:

1. Add a delay between creating the issue and starting the PR creation (not good IMO)
2. Remove the filtering on pulumiupgradeproviderissue so we just fetch the 100 latest issues in the repo and look at their titles for matching the exact upgrade issue title format
3. Try switching away from a keyword in the body and see if labels indexing is faster. This would require setting up new labels in all repos and reworking more aspects.

PR #293, opened 2024-11-08
Stop filtering on search token when finding applicable issues

danielrbradley:
This might be running very quickly after the issue was created and it might not yet be indexed. Instead, just list the latest 100 issues and filter on exact title format instead.

The current rate of issue creation is low enough that the latest 100 issues should safely include all open upgrade issues.

Fixes https://github.com/pulumi/upgrade-provider/issues/291

We'll continue to use the token for issue de-duplication as this only happens once per day and so is reliably in the GH issue search index.

### DIFF
--- a/upgrade/steps_helpers.go
+++ b/upgrade/steps_helpers.go
@@ -620,7 +620,6 @@ var getExpectedTargetFromIssues = stepv2.Func11E("From Issues", func(ctx context
 		"--state=open",
 		"--repo="+name,
 		"--limit=100",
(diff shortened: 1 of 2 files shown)


danielrbradley:
Test run using this pre-release version: https://github.com/pulumi/pulumi-xyz/actions/runs/11742090038/job/32712100051

Resulting PR has correct "fixes" directive: https://github.com/pulumi/pulumi-xyz/pull/518

PR #306, opened 2025-02-06
Fix upstream.sh path

danielrbradley:
In https://github.com/pulumi/ci-mgmt/pull/1362 we're moving the location of the upstream.sh into the scripts directory to keep the root of the repo cleaner. This update keeps this tool in line with the new location.

### DIFF
--- a/upgrade/steps.go
+++ b/upgrade/steps.go
@@ -163,11 +163,11 @@ func UpgradeProviderVersion(
 			// We need to remove any patches to so we can cleanly pull the next upstream version.
 			step.Cmd("git", "reset", "HEAD", "--hard").In(&upstreamDir),
 			// Load patches into a branch.
-			step.Cmd("./upstream.sh", "checkout").In(&repo.root),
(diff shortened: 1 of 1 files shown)


upgrade/steps.go
```diff
@@ -163,11 +163,11 @@ func UpgradeProviderVersion(
 			// We need to remove any patches to so we can cleanly pull the next upstream version.
 			step.Cmd("git", "reset", "HEAD", "--hard").In(&upstreamDir),
 			// Load patches into a branch.
-			step.Cmd("./upstream.sh", "checkout").In(&repo.root),
+			step.Cmd(".scripts/upstream.sh", "checkout").In(&repo.root),
```
thomas11, line 166: missing slash?

danielrbradley: Yup! Good catch! Guess the tests don't catch much in this project 🙈

danielrbradley:
I've opened https://github.com/pulumi/upgrade-provider/issues/307 to track better testing here.

Issue #305, opened 2025-02-04
upgrade-provider should create a unique PR


corymhall:
### What happened?

It looks like when we run the upgrade bridge downstream tests it will overwrite any existing branch/PR if the bridge version is the same. For example, this PR was originally generated by the upgrade-bridge workflow to upgrade the bridge to v3.102.0. The branch was then overwritten and the PR title and description was updated

https://github.com/pulumi/pulumi-aws/pull/5160

One possible solution is to add the `pr-title-prefix` to the branch name

### Example

https://github.com/pulumi/pulumi-aws/pull/5160

PR was first generated with

```
upgrade-provider pulumi/pulumi-aws --kind=bridge --pr-reviewers=VenelinMartinov
```

And then it was overwritten by this

```
upgrade-provider pulumi/pulumi-aws --kind=bridge --pr-reviewers=guineveresaenger --pr-title-prefix=[DOWNSTREAM TEST][PLATFORM] --target-pulumi-version=2c86a1fd6e601178468f2b8229933f2b8f4667a8
```

### Output of `pulumi about`

NA

### Additional context

This issue could be potentially dangerous because you can approve and enable auto-merge on a PR and then have it updated to an experimental commit.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #308, opened 2025-02-06
add PR prefix to branch name

VenelinMartinov:
This PR adds the PR prefix to the branch name when running upgrade-provider. This should avoid the confusing and dangerous situation where legitimate upgrade branches are clobbered by test runs of upgrade-provider.

fixes https://github.com/pulumi/upgrade-provider/issues/305

### DIFF
--- a/upgrade/steps_test.go
+++ b/upgrade/steps_test.go
@@ -55,6 +55,16 @@ func TestGetWorkingBranch(t *testing.T) {
 			branchSuffix:        "foo",
 			expected:            "upgrade-pulumi-terraform-bridge-to-1.2.3-foo",
 		},
+		{
+			c: Context{
+				UpgradeBridgeVersion: true,
+				TargetBridgeRef:      &Version{SemVer: semver.MustParse("v1.2.3")},
+				PRTitlePrefix:        "[DOWNSTREAM TEST] [PLATFORM]",
+			},
+			targetBridgeVersion: &Version{SemVer: semver.MustParse("1.2.3")},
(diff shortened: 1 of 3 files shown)


upgrade/steps_test.go
```diff
@@ -39,9 +41,20 @@ func TestGetWorkingBranch(t *testing.T) {
...
 			},
+			branchSuffix:        "",
 			targetBridgeVersion: &Version{SemVer: semver.MustParse("1.2.3")},
 			expected:            "upgrade-pulumi-terraform-bridge-to-1.2.3",
 		},
+		{
+			c: Context{
+				UpgradeBridgeVersion: true,
+				TargetBridgeRef:      &Version{SemVer: semver.MustParse("v1.2.3")},
+				PRTitlePrefix:        "foo",
```
t0yv0, line 52: What is a realistic PR title prefix? What is a PR title prefix?

t0yv0: This is here right? 

https://github.com/pulumi/upgrade-provider/blob/03d6e987a846ff0b8d848f02a0abd1476aee4efd/main.go#L256

t0yv0: Can you put some checks for real values.
```
""
"[TEST]"

```
Presumably `[DOWNSTREAM TEST][PLATFORM]` is a prefix too? 

https://github.com/pulumi/pulumi-aws/pull/5169

Just to make sure it's sane.

t0yv0 (approved):
Thanks for jumping on this quickly! This is going to certainly save us some confusion.

PR #311, opened 2025-04-21
Add patch release option

VenelinMartinov:
This PR adds a --patch-release option which adds the `needs-release/patch` label to the PR generated by `upgrade-provider`.

Example: https://github.com/pulumi/pulumi-random/pull/1791

Part of https://github.com/pulumi/pulumi-terraform-bridge/issues/2267


VenelinMartinov:
I think I don't actually need this as this can be solved entirely in upgrade-provider-action

PR #312, opened 2025-05-01
Generate, do not build the SDKs

t0yv0:
The change swaps out `make build_sdk` with `make generate_sdk` which is lighter weight and is less prone to failure. 

In particular if upgrade succeeded partially but possibly the .NET SDK does not build would rather have that failing PR stood up than fail and abort having a PR out.

Also it is lighter on resources and imposes fewer constraints on the environment it executes in.

Problem sparking this: https://github.com/pulumi/pulumi-alicloud/issues/990

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,10 @@ Flags:
       --repo-path string                Clone the provider repo to the specified path.
       --target-bridge-version ref       The desired bridge version to upgrade to. Git hash references permitted. (default <latest>)
       --target-pulumi-version ref       Upgrade the provider to the passed pulumi/{pkg,sdk} version.
-                                        
+
                                         If no version is passed, the pulumi/{pkg,sdk} version will track the bridge
       --target-version string           Upgrade the provider to the passed version.
-                                        
+
                                         If the passed version does not exist, an error is signaled.
       --upstream-provider-name string   The name of the upstream provider.
                                         Required unless running from provider root and set in upgrade-config.yml.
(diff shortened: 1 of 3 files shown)


blampe (approved):
I think this would fix a big annoyance of mine where this job fails without putting a PR up, yay!

iwahbe (approved):

t0yv0:
>  I think this would fix a big annoyance of mine where this job fails without putting a PR up, yay!

I do not think it gets us quite there yet, I'll dig a bit further into this if time permits, (1) something failed but made a green job so we drop the error somewhere; and (2) the entire stack should indeed tolerate errors in terms of putting up a PR; it should save any WIP it's done and put up the PR, logging the error on the PR, that'd be swell 

Let's ship this one though.

Eyal-Shalev:
Hi, this broke our build.

We are using `pulumi/pulumi-upgrade-provider-action` and with no way of specifying the `pulumi/upgrade-provider` package our builds started breaking because you changed the command from `build_sdks` to `generate_sdks`.

The minimum that is required for breaking changes is to write something in the main Readme.md file. Better yet is to fallback on the previous command and write a message asking to change it.

Eyal-Shalev:
The only output from the command was:

```
---- Update Repository ----
- Update TF Provider
 Lookup Tag SHA: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 /opt/hostedtoolcache/go/1.21.13/x64/bin/go get github.com/port-labs/terraform-pr...: done
error: exit status 2:
make: *** No rule to make target 'generate_sdks'.  Stop.
Error: Process completed with exit code 1.
```

t0yv0:
@Eyal-Shalev sorry about your breaking build! You can alias build_sdk to generate_sdk to compensate for the moment. You are making a good call-out and that is how "general availability" software at Pulumi is supported (versions 1.0+); unfortunately unless I am mistaken we are not at a place where upgrade-provider is ready to offer these guarantees. We're co-evolving it with `ci-mgmt` repo which defines `generate_sdk` on recent versions.  I am sorry if README was not clear enough on the current status. I will double check with the team what our plans are here and if I'm misreading the support level, or how can we package these things in the future that can be actually relied on and versioned properly.

Eyal-Shalev:
Note that if https://github.com/pulumi/pulumi-upgrade-provider-action provided a way to specify the version of https://github.com/pulumi/upgrade-provider it would allow us to lock a specific working version.

This could be a middleground if this project API is not stable.

t0yv0:
Great suggestion, having a look!

PR #313, opened 2025-05-02
Ignore errors from make

t0yv0:
Following up on https://github.com/pulumi/upgrade-provider/pull/312

At Pulumi when using with GitHub actions we almost certainly would prefer the tool to check in a WIP in-progress PR that failed rather than fail and check in no PR. Perhaps some of the actions it took are valid just require manual intervention to post-fix-up.

Ideally the control flow would be: 

```
try:
  make tfgen
  make generate_sdks   
finally:
  create-pr
```

However I do not think the workflow monad here permits a try-finally construct that I could implement quickly. 

As an alternative we can ignore errors from `make tfgen` and `make generate_sdks`.

As another alternative we can make ignoring these errors a flag. Admittedly it is a very strange default behavior.

### DIFF
--- a/step/v2/util.go
+++ b/step/v2/util.go
@@ -6,6 +6,7 @@ import (
 	"io/fs"
 	"os"
 	"os/exec"
+	"strings"
 )
 
 // Name a value so it is viable to the user.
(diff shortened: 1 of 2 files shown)


t0yv0:
Sorry I am a little confused - what do you mean starting from scratch?

iwahbe:
> Sorry I am a little confused - what do you mean starting from scratch?

If I saw that a PR failed to create, I would go `cd pulumi/pulumi-${that-provider} && upgrade-provider ...`.

PR #320, opened 2025-06-30
Do not default to upgrading java version in all ugprade jobs

VenelinMartinov:
Java upgrades should not be mixed with provider upgrades.

### DIFF
--- a/main.go
+++ b/main.go
@@ -169,7 +169,7 @@ func cmd() *cobra.Command {
 
 	pulumiDev, _ := strconv.ParseBool(os.Getenv("PULUMI_DEV"))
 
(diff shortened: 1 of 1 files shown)


VenelinMartinov:
This change is part of the following stack:

- #320 ◀
    - #321
        - #322

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

t0yv0 (approved):

t0yv0:
What replacement job will be doing java upgrades then? Are we tracking?

VenelinMartinov:
Opened https://github.com/pulumi/ci-mgmt/issues/1605 to track this

PR #321, opened 2025-06-30
Remove logic for ugprading PF since that does not exist any more

VenelinMartinov:
We deleted the /pf module in the bridge so no need to keep the logic around here.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,13 +1,31 @@
+version: "2"
 linters:
+  exclusions:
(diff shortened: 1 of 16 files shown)


VenelinMartinov:
This change is part of the following stack:

- #320
    - #321 ◀
        - #322

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

t0yv0 (approved):

PR #327, opened 2025-07-01
Remove logic for current directory - it should always be the provider repo

VenelinMartinov:
Removes the logic for using the tool when not inside the provider repo - we don't use it for our workflows and it likely doesn't work.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,7 +38,6 @@ func cmd() *cobra.Command {
 	var upgradeKind []string
 	var repoName string
 	var repoOrg string
-	var repoPath string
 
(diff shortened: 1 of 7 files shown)


ringods:
@t0yv0 @VenelinMartinov as long as the Github `upgrade-provider-action`, part of the `upgrade-provider` Github workflow remains working, fine for me.

https://github.com/pulumiverse/pulumi-grafana/blob/96ed13ef40d43494bfe569295aa794967260b5ea/.github/workflows/upgrade-provider.yml#L74

PR #328, opened 2025-07-01
Separate out logic for checking upstream version

VenelinMartinov:
`upgrade-provider` is actually used in two ways:
1. As a tool to check the upstream repo for new versions and open a corresponding GH issue (`check-upstream-version`)
2. As a tool to upgrade the provider

Previously the logic for the two uses was intertwined making reasoning about consequences more difficult. This PR separates them into two function calls.

### DIFF
--- a/upgrade/check_upstream.go
+++ b/upgrade/check_upstream.go
@@ -1,0 +1,66 @@
+package upgrade
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/pulumi/upgrade-provider/colorize"
+	stepv2 "github.com/pulumi/upgrade-provider/step/v2"
+)
+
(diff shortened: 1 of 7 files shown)


upgrade/check_upstream.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	})
+	if err != nil {
+		return err
+	}
+
+	if GetContext(ctx).UpgradeProviderVersion {
+		pipelineName := fmt.Sprintf("New upstream version detected: v%s", upgradeTarget.Version)
+		return stepv2.PipelineCtx(ctx, pipelineName, func(ctx context.Context) {
+			createUpstreamUpgradeIssue(ctx,
```
t0yv0, line 55: Why is this editing issues? This is still quite tangled. 

At the very least the name CheckUpstream should reflect everything this is trying to do.

VenelinMartinov:

PR #329, opened 2025-07-01
Remove logic for inferring current upstream version - use a file to record it instead

VenelinMartinov:
This PR removes the logic for inferring the current upstream version. That logic adds unnecessary complexity as the repo should know that information.

Instead of that we will now record the version in the `.upgrade-config.yml` file in each provider repo. That will be bumped each time we make an upgrade by `upgrade-provider`.

### DIFF
--- a/upgrade/upgrade_provider.go
+++ b/upgrade/upgrade_provider.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
+
 	"github.com/pulumi/upgrade-provider/colorize"
 	"github.com/pulumi/upgrade-provider/step"
 	stepv2 "github.com/pulumi/upgrade-provider/step/v2"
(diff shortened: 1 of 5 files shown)


t0yv0:
So the argument here is that ugprades always target a partricular verison, and we'd just use `.upgrade-config.yml` as "memory" to save it, instead of trying to infer this. I guess this makes sense to me! I am always wary of new config options but this seems to make sense.

upgrade/upgrade_provider.go
```diff
@@ -298,3 +306,23 @@ func tfgenAndBuildSDKs(
...
+		content := stepv2.ReadFile(ctx, configFile)
+		lines := strings.Split(content, "\n")
+		var newLines []string
+		for _, line := range lines {
+			if !strings.Contains(line, "current-upstream-version:") {
+				newLines = append(newLines, line)
+			}
+		}
+		content = strings.Join(newLines, "\n")
+		content += "current-upstream-version: " + version.String() + "\n"
```
t0yv0, line 321: Is not this file in YAML? Can we write a struct for it, document the fields and read/write using YAML tooling instead of munging text?

VenelinMartinov:

PR #330, opened 2025-07-01
remove logic for inferring upstream org

VenelinMartinov:
That is unnecessary and can be replaced by a config in `upgrade-config.yml`. This also makes it easier to change the configuration when upstream repos move between organizations.


t0yv0:
Is it different from 

https://github.com/pulumi/pulumi-cloudflare/blob/8cd6c4c8857a8e0f4dcfdc3c019076e04924376e/provider/resources.go#L78

VenelinMartinov:

PR #332, opened 2025-07-01
Remove target version inference

VenelinMartinov:
This removes the logic for inferring the target version - we have not been using that for quite a while as we now have a separate job which infers the target version - `check-upstream-version`.


t0yv0 (approved):

VenelinMartinov:

PR #336, opened 2025-07-02
End-to-end tests

VenelinMartinov:
This adds end-to-end tests for `upgrade-provider` without any test recordings or mocking.


upgrade_test.go
```diff
@@ -0,0 +1,213 @@
...
+	upgradeConfig, err := os.ReadFile(filepath.Join(folder, ".upgrade-config.yml"))
+	require.NoError(t, err)
+
+	require.Contains(t, string(upgradeConfig), fmt.Sprintf("current-upstream-version: %s", version))
+	// TODO: assert on upstream submodule sha
+}
+
+func TestGCPBridgeUpgrade(t *testing.T) {
+	folder := tempDir(t, "pulumi-gcp")
+	version := "bfbfac3b8a54f936f55f86cddf91a1b2184bc7b9"
```
t0yv0, line 169: Interesting, at some point this will need to decouple from gcp but I like how you are pinning a commit so it is probably reasonable to start reusing that as a test case until the coupling issues actually arise.

t0yv0 (approved):

VenelinMartinov:

PR #372, opened 2026-05-01
Add allow-major provider upgrade flag

corymhall:
## Summary
- add an `--allow-major` flag for automation that permits major provider upgrades only when the planned target actually crosses a major boundary
- keep `--major` as a strict assertion that errors on non-major upgrades
- document `allow-major` in provider config and cover config binding plus major policy behavior in tests

## Motivation

We want the ability to opt-in certain providers to major version upgrades. https://github.com/pulumi/pulumi-fastly/pull/1009 is an example of where this would be useful. Providers will now be able to opt-in to this ability by updating `.upgrade-config.yml` to include `allow-major: true`.

re https://github.com/pulumi/home/issues/4562

## Testing
- `go test ./...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Usage:
 Flags:
       --allow-missing-docs              If true, don't error on missing docs during tfgen.
                                         This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.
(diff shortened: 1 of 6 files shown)


guineveresaenger (approved):
SGTM

PR #374, opened 2026-05-07
Fix mise env refresh after provider updates

corymhall:
## Summary
- persist the Pulumi and Go mise env vars read from the upgraded provider go.mod into the shared step context
- run `mise env --json` and downstream generation commands under those refreshed vars instead of stale job-level values
- add a regression test that starts with stale mise vars and verifies `mise upgrade`, `mise env`, and a downstream command all see the upgraded versions

## Test
- `go test ./...`

### DIFF
--- a/upgrade/mise_test.go
+++ b/upgrade/mise_test.go
@@ -1,0 +1,83 @@
+package upgrade
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	stepv2 "github.com/pulumi/upgrade-provider/step/v2"
+)
+
+func TestRunMiseUpgradeRefreshesFromUpdatedProviderGoMod(t *testing.T) {
+	t.Setenv("PULUMI_VERSION_MISE", "3.228.0")
+	t.Setenv("GO_VERSION_MISE", "1.25.8")
+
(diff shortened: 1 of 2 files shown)


upgrade/mise_test.go
```diff
@@ -0,0 +1,83 @@
...
+		miseEnvVars := map[string]*stepv2.EnvVar{}
+
+		var ok bool
+		ctx, ok = runMiseUpgrade(ctx, repo, &env, miseEnvVars)
+		if !ok {
+			stepv2.HaltOnError(ctx, errors.New("mise not available"))
+		}
+		ctx = refreshMiseEnv(ctx, &env, miseEnvVars)
+
+		stepv2.Cmd(ctx, "sh", "-c", fmt.Sprintf(
```
pose: Are we sure that `${PULUMI_VERSION_MISE:-}` is supported in `sh` and we don't need to use `bash` explicitely here? Try to avoid a "works on my computer" situation of `sh` being aliased to `bash`.

pose (approved):

upgrade/mise_test.go
corymhall: Good call to make the test explicit. `${VAR:-}` is POSIX shell syntax, so the original `sh -c` was not relying on bash specifically, but this test already uses bash for the fake `mise` script. I switched this downstream probe to `bash -c` as well to remove the ambiguity.

PR #381, opened 2026-07-09
cli: show default flag values when calling --help

### LINKED ISSUE #377: cli: show default flag values when calling `--help`
## Acceptance Criteria

Example (find the proper flag values):

```
> $ upgrade-provider --help
...
Flags:
      --allow-major=false                     Allow the provider to upgrade to a new major version when one is available.
      --allow-missing-docs=true              If true, don't error on missing docs during tfgen.
                                        This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.
      --dry-run=false                         If true, don't actually create any PRs
      --kind strings=all                    The kind of upgrade to perform:

                                        - "all": Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider".
                                        - "bridge": Upgrade the bridge only.
                                        - "provider": Upgrade the upstream provider only. (default [all])
      --major                           Upgrade the provider to a new major version.
      --pr-assign string                A user to assign the upgrade PR to.
      --pr-description string           Extra text to insert in the generated pull request description.
...
```

pose:
## Summary

Closes #377.

`upgrade-provider --help` previously hid a flag's default value whenever it was the "zero value" for its type (e.g. `false` for booleans, `""` for strings). This made it impossible to tell, for example, whether `--allow-missing-docs` or `--dry-run` default to `true` or `false` without reading the source.

This PR customizes cobra's usage template so flag defaults are always shown, inline right after the flag name (and type, if any), e.g.:

```
Flags:
      --allow-major=false                    Allow the provider to upgrade to a new major version when one is available.
      --allow-missing-docs=false             If true, don't error on missing docs during tfgen.
                                             This is equivalent to setting PULUMI_MISSING_DOCS_ERROR=${! VALUE}.
      --dry-run=false                        If true, don't actually create any PRs
  -h, --help                                 help for upgrade-provider
      --kind strings=all                     The kind of upgrade to perform:
      ...
      --major=false                          Upgrade the provider to a new major version.
      --pr-assign string                     A user to assign the upgrade PR to.
      ...
      --target-bridge-version ref=<latest>   The desired bridge version to upgrade to. Git hash references permitted.
```

Flags whose default is genuinely empty (e.g. `--pr-assign string`) are left without a `=...` suffix, matching the acceptance criteria in the issue. Cobra's own implicit `--help`/`--version` flags are left untouched since their default is never meaningful.

## Changes

- `helpflags.go` (new): implements `installDefaultValueHelp`, a replacement for pflag's `FlagUsages` that always renders a flag's default value (even zero values) inline as `--flag=default`, and wires it into cobra's usage template.
- `main.go`: calls `installDefaultValueHelp(cmd)` after all flags are registered.
- `main_test.go`: adds `TestHelpShowsDefaultFlagValues`, which asserts boolean flags show `=false`, non-zero defaults for string/slice flags are shown, empty-default flags are not, and cobra's own `--help` flag is unaffected.
- `README.md`: updates the example `--help` output (which had also drifted out of sync with the current flag set) to match the new format.

## Testing

- `go build ./...`
- `go test ./...` — all tests pass, including the new `TestHelpShowsDefaultFlagValues`.
- Manually ran `upgrade-provider --help` and `PULUMI_DEV=1 upgrade-provider --help` to confirm formatting, alignment, and hidden-flag behavior are correct.

---
*Created with [Eon](https://eon.corp.pulumi.com/session/b85da2691cc14da4487d1792b4dd94a1)*


helpflags.go
```diff
@@ -0,0 +1,176 @@
+package main
```
pose: @pulumi-eon could we simplify this file even further? Is there a built-in alternative in Cobra that still show us all the defaults when printing usage but doesn't require this customization?

corymhall (commented):
What if we kept Cobra’s standard formatter and handling and add just the missing boolean defaults instead?

For the current flag set, Cobra already displays the non-zero defaults, the information it suppresses is the false default on the boolean flags. We could surface that with a small `BoolVar` helper that appends `(default: false)` to the usage text, deriving it from the same default value to prevent drift.

That would change the presentation from `--dry-run=false` to something like `--dry-run ... (default: false)`, but it would avoid maintaining a custom copy of Cobra/pflag’s formatting logic.

e.g.

```go
func boolFlag(
    flags *pflag.FlagSet,
    target *bool,
    name string,
    defaultValue bool,
    usage string,
) {
    flags.BoolVar(
        target,
        name,
        defaultValue,
        fmt.Sprintf("%s (default: %t)", usage, defaultValue),
    )
}

boolFlag(
    cmd.PersistentFlags(),
    &context.DryRun,
    "dry-run",
    false,
    "If true, don't actually create any PRs.",
)
```

pose:
> What if we kept Cobra’s standard formatter and handling and add just the missing boolean defaults instead?

That's music to my ears. Let's me do that!

corymhall (approved):

pose:
@pulumi-eon open a new PR since we removed "kind -> check-upstream-version" from the README and we would like to be present.

PR #383, opened 2026-07-13
Propagate Git identity to patched provider upgrades

corymhall:
## Summary

- resolve Git author and committer identity from explicit `GIT_AUTHOR_*` / `GIT_COMMITTER_*` values, then effective provider-repository config
- propagate the resolved identity to `scripts/upstream.sh`, Git commands inside the upstream submodule, and subsequent upgrade commands without writing global config
- fail with repository-local configuration instructions when identity remains incomplete
- document and test the patched-provider behavior, including preservation of patch authors during `git am`

## Why

Repository-local `user.name` and `user.email` in the provider root do not apply inside the `upstream` submodule. In sandboxed environments without global Git config, `scripts/upstream.sh checkout` therefore fails when `git am` attempts to create commits.

The preflight is intentionally limited to patched provider-version upgrades immediately before entering the upstream patch workflow. It does not depend on `gh api user` or broader checkout timing changes.

## Testing

- `go test -count=1 ./...`
- `go vet ./...`
- `golangci-lint run --new-from-rev=HEAD`
- `go test -race ./step/... ./upgrade/...`

### DIFF
--- a/step/step.go
+++ b/step/step.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


step/step.go
```diff
@@ -34,6 +35,50 @@ type step struct {
...
+			merged[key] = value
+		}
+	}
+	for key, value := range values {
+		merged[key] = value
+	}
+	return context.WithValue(ctx, commandEnvKey{}, merged)
+}
+
+func commandEnv(ctx context.Context) []string {
```
pose, line 55: (offtopic)I am baffled there's no better way to do this!

pose (approved):

step/step.go
corymhall: Yeah I was surprised by this too! I guess most people just set it globally and don't have to worry about it.

